PR #1, opened 2018-07-03
Basic Wpf UserControl

Ch4rg3r:
Added Wpf UserControl with Autosizing using WindowsFormsHost


LibVLCSharp.Wpf/LibVLCSharp.Wpf.csproj
```diff
@@ -0,0 +1,67 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<Project ToolsVersion="15.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <Import Project="$(MSBuildExtensionsPath)\$(MSBuildToolsVersion)\Microsoft.Common.props" Condition="Exists('$(MSBuildExtensionsPath)\$(MSBuildToolsVersion)\Microsoft.Common.props')" />
+  <PropertyGroup>
+    <Configuration Condition=" '$(Configuration)' == '' ">Debug</Configuration>
+    <Platform Condition=" '$(Platform)' == '' ">AnyCPU</Platform>
+    <ProjectGuid>{DA7A2677-0944-481F-A59B-9128FC54FD5F}</ProjectGuid>
+    <OutputType>Library</OutputType>
+    <AppDesignerFolder>Properties</AppDesignerFolder>
+    <RootNamespace>LibVLCSharp.Wpf</RootNamespace>
```
mfkl: Can you rename `LibVLCSharp.Wpf` to `LibVLCSharp.WPF` everywhere please?

LibVLCSharp.Wpf/VLCViewControl.xaml.cs
```diff
@@ -0,0 +1,72 @@
...
+using System.Windows.Documents;
+using System.Windows.Input;
+using System.Windows.Media;
+using System.Windows.Media.Imaging;
+using System.Windows.Navigation;
+using System.Windows.Shapes;
+
+namespace LibVLCSharp.Wpf
+{
+    public partial class VLCViewControl : UserControl, LibVLCSharp.Shared.IVideoView
```
mfkl: To be consistent with the other platforms, can you rename `VLCViewControl` with `VideoView`?

mfkl: Also, I believe your control needs to be `IDisposable` and dispose of libvlc and mediaplayer in the dispose method

LibVlcSharp.Wpf.Sample/App.xaml
```diff
@@ -0,0 +1,9 @@
+﻿<Application x:Class="LibVlcSharp.Wpf.Sample.App"
```
mfkl: rename `LibVlcSharp.Wpf.Sample` to `LibVLCSharp.WPF.Sample` everywhere please.

LibVlcSharp.Wpf.Sample/App.xaml.cs
```diff
@@ -0,0 +1,17 @@
...
+using System.Configuration;
+using System.Data;
+using System.Linq;
+using System.Threading.Tasks;
+using System.Windows;
+
+namespace LibVlcSharp.Wpf.Sample
+{
+    /// <summary>
+    /// Interaktionslogik für "App.xaml"
```
mfkl: this comment looks very german :D

jeremyVignelles:
I really wonder if making a wpf control is a good thing. Look at what happened in Vlc.DotNet :
We had this kind of "WPF" control, which is just a wrapper over WinForms. We had airspace issues and user complained about that. Some migrated to Meta.Vlc, which provided a more WPF-ish way of doing that : an ImageSource. With that, you could use it like a Wpf image, use it everywhere like brushes...
I decided to replace the implementation of the WPF control which I considered useless
 (Wrapping a component in a WinFormsHost is really easy). I quickly realized that implementing that with an ImageSource would come at a terrible cost : perf is bad...

Here's the situation now, in Vlc.DotNet :
- User have the choice of their technology (which is good)
- User tend to install the Wpf control by default, despite my numerous warnings that it comes at the cost of performance, and then complain that the perfs are poor when they use 50 streams.

For the full-length story, see https://github.com/ZeBobo5/Vlc.DotNet/issues/249

With that experience, I would think twice before doing a "Wpf" control. 
In Vlc.DotNet, we were limited to windows-only desktop apps, so not providing a WPF option was possible, but now that we should support Xamarin.Forms with a Wpf backend and Eto.Forms...

What would you think of having several Wpf implementations?
Like :
.Wpf.WinForms/.Xamarin.Forms.Wpf.WinForms
and 
.Wpf.ImageSource/.Xamarin.Forms.Wpf.ImageSource

Users would have to read a short readme to take an enlighted decision, but if that could avoid the issues about perfs/about airspace.

LibVLCSharp.Wpf/LibVLCSharp.Wpf.csproj
jeremyVignelles: I tend to wirte Wpf, to case all my namespaces and classes the C# way ( I have tools that complains when I don't)

mfkl:
Thanks for the input @jeremyVignelles 

Unity's dude's take on WPF/D3D situation: https://github.com/sharpdx/SharpDX/issues/599#issuecomment-114016945 TL;DR: "D3DImage is fundamentally broken and unsuitable for efficient D3D rendering in WPF."

So right now the choices are:
- WinForms Host -> Airspace issues.
- WPF -> bad performance (copy).
- Just Not support WPF -> 1 less target for Xamarin.Forms.
- Patch vlc/libvlc to provide a new API which would help bypass all those issues/bad components? Not sure that's even possible /cc @jbkempf

jbkempf:
You have no choice here, if you want to target most platforms: Winforms -> airspacing issues.

D3DImage does too many copies, and DXGI integration is not working on Windows 7, IIRC, unless people have D3D11 working, which is rare.

jeremyVignelles:
If I remember correctly, WPF is DirectX 9 only. Microsoft provides a
D3D11Image, but it looks like an emulation layer over the D3DImage built
into WPF

Ch4rg3r:
I know the way VLC.DotNet does this. I thought about DirectX and OpenGL too, but it seems they somehow require Handles too. Even Microsoft wasn't able to fix this - a fix was availible for a short time in PresentationFramework 4.5

jbkempf:
@Ch4rg3r ok, great.

My opinion is to use that WPF/Handle and try to fix the airspacing issues.

Ch4rg3r:
@jbkempf i guess that's true.  The mentioned way on stackoverflow (no idea if it works) draws the WPF controls on Win32 on top of the WindowsFormsHost and forwards mouse actions to the WPF controls. This way the WPF is copied and drawn twice. Guess that has less performance impact

jeremyVignelles:
I agree with j-b : let's go for the perfs.
For the airspace issue, I think we can try the way that was suggested in Vlc.DotNet : https://github.com/kolorowezworki/Airhack
Thanks @Ch4rg3r for your time and reactivity!

Ch4rg3r:
@jeremyVignelles seems to be the implementation of what i meant, but simpler. And i just tested it. It works, enabling mouse actions for both background and foreground.

jeremyVignelles:
How do you see the interaction with this project? I thought that this would be something to "recommend" for users facing this issue, but not something embedded in the library since most of the users, I guess, just want to display the video, not to hover things on top of that.

However I don't know the status of this project (licensing, packaging, support...), and maybe @kolorowezworki can answer that?

Ch4rg3r:
I would make VideoView from it, using the background as MediaPlayer, giving the User of this library the ability to add child elements to the foreground. I guess it requires a few changes

mfkl:
@Ch4rg3r Let us know when you think this is good for review/testing.

Ch4rg3r:
@mfkl the last change works for me. If you can test it's functionality, please do it. Maybe you find issues or things to be improved.

LibVLCSharp.WPF/ForegroundWindow.xaml.cs
```diff
@@ -0,0 +1,106 @@
...
+using System.Windows.Media.Imaging;
+using System.Windows.Navigation;
+using System.Windows.Shapes;
+
+namespace LibVLCSharp.WPF
+{
+    internal partial class ForegroundWindow : Window
+    {
+        private Window wndhost;
+        private ContentControl bckgnd;
```
mfkl: Use "_" consistently for private fields please.

LibVLCSharp.WPF/ForegroundWindow.xaml.cs
```diff
@@ -0,0 +1,106 @@
...
+        private void Bckgnd_Loaded(object sender, RoutedEventArgs e)
+        {
+            wndhost = Window.GetWindow(bckgnd);
+            this.Owner = wndhost;
+            wndhost.Closing += Wndhost_Closing;
+            wndhost.SizeChanged += Wndhost_SizeChanged;
+            wndhost.LocationChanged += Wndhost_LocationChanged;
+            try
+            {
+                Point locationFromScreen = bckgnd.PointToScreen(new Point(0, 0));
```
mfkl: This `new Point(0, 0)` allocation can be cached in a readonly field.

LibVLCSharp.WPF/ForegroundWindow.xaml.cs
```diff
@@ -0,0 +1,106 @@
...
+                PresentationSource source = PresentationSource.FromVisual(wndhost);
+                System.Windows.Point targetPoints = source.CompositionTarget.TransformFromDevice.Transform(locationFromScreen);
+                this.Left = targetPoints.X;
+                this.Top = targetPoints.Y;
+                Vector size = bckgnd.PointToScreen(new Point(bckgnd.ActualWidth, bckgnd.ActualHeight)) - bckgnd.PointToScreen(new Point(0, 0));
```
mfkl: This `new Point(0, 0)` allocation can be cached in a readonly field.

LibVLCSharp.WPF/ForegroundWindow.xaml.cs
```diff
@@ -0,0 +1,106 @@
...
+                this.Height = size.Y;
+                this.Width = size.X;
+                this.Show();
+                wndhost.Focus();
+            }
+            catch
+            {
+                this.Hide();
```
mfkl: Maybe log an error or something? What could possibly happen here?

LibVLCSharp.WPF/ForegroundWindow.xaml.cs
```diff
@@ -0,0 +1,106 @@
...
+            }
+        }
+
+        private void Wndhost_LocationChanged(object sender, EventArgs e)
+        {
+            Point locationFromScreen = bckgnd.PointToScreen(new Point(0, 0));
```
mfkl: This `new Point(0, 0)` allocation can be cached in a readonly field.

LibVLCSharp.WPF/ForegroundWindow.xaml.cs
```diff
@@ -0,0 +1,106 @@
...
+            PresentationSource source = PresentationSource.FromVisual(wndhost);
+            System.Windows.Point targetPoints = source.CompositionTarget.TransformFromDevice.Transform(locationFromScreen);
+            this.Left = targetPoints.X;
+            this.Top = targetPoints.Y;
+        }
+
+        private void Wndhost_SizeChanged(object sender, SizeChangedEventArgs e)
+        {
+            Point locationFromScreen = bckgnd.PointToScreen(new Point(0, 0));
```
mfkl: This `new Point(0, 0)` allocation can be cached in a readonly field.

LibVLCSharp.WPF/ForegroundWindow.xaml.cs
```diff
@@ -0,0 +1,106 @@
...
+            PresentationSource source = PresentationSource.FromVisual(wndhost);
+            System.Windows.Point targetPoints = source.CompositionTarget.TransformFromDevice.Transform(locationFromScreen);
+            this.Left = targetPoints.X;
+            this.Top = targetPoints.Y;
+            Vector size = bckgnd.PointToScreen(new Point(bckgnd.ActualWidth, bckgnd.ActualHeight)) - bckgnd.PointToScreen(new Point(0, 0));
```
mfkl: This `new Point(0, 0)` allocation can be cached in a readonly field.

LibVLCSharp.WPF/VideoView.cs
```diff
@@ -0,0 +1,99 @@
...
+
+namespace LibVLCSharp.WPF
+{
+    public partial class VideoView : UserControl, LibVLCSharp.Shared.IVideoView, IDisposable
+    {
+        private LibVLCSharp.Shared.MediaPlayer mediaPlayer;
+        private LibVLCSharp.Shared.LibVLC libVLC;
+        private double controlWidth;
+        private double controlHeight;
+        private ForegroundWindow foreground;
```
mfkl: Same here, please use "_" for consistent code styling ;-)

LibVLCSharp.WPF/VideoView.cs
```diff
@@ -0,0 +1,99 @@
...
+                VLCResize();
+            }
+        }
+
+        public void Dispose()
+        {
+            if (mediaPlayer.IsPlaying)
+                mediaPlayer.Stop();
+
+            mediaPlayer.Dispose();
```
mfkl: You may add `_mediaPlayer.Hwnd = IntPtr.Zero`

LibVLCSharp.WPF/VideoView.cs
```diff
@@ -0,0 +1,99 @@
...
+            ResourceDictionary res = Application.LoadComponent(new Uri("/LibVLCSharp.WPF;component/Styles/VideoView.xaml", UriKind.RelativeOrAbsolute)) as ResourceDictionary;
+            this.Style = res["VideoViewStyle"] as Style;
+
+            foreground = new ForegroundWindow(this);
+
+            Shared.Core.Initialize();
+            libVLC = new Shared.LibVLC();
+            mediaPlayer = new Shared.MediaPlayer(libVLC);
+
+            this.SizeChanged += OnSizeChanged;
```
mfkl: I think this needs to be unregistered in the Dispose method.

LibVLCSharp.WPF/VideoView.cs
```diff
@@ -0,0 +1,99 @@
...
+
+            controlHeight = this.Height;
+            controlWidth = this.Width;
+
+            this.Loaded += VideoView_Loaded;
```
mfkl: I think this needs to be unregistered in the Dispose method.

LibVLCSharp/Shared/MediaPlayer.cs
```diff
@@ -631,6 +631,11 @@ struct Native
 #endif
         }
 
+        public void SetVideoFormatCallbacks()
+        {
+            throw new NotImplementedException();
```
mfkl: What's up with this? :D https://github.com/Ch4rg3r/libvlcsharp/blob/7ddc025392caf39ec1277c4f59585b672d88cf8b/LibVLCSharp/Shared/MediaPlayer.cs#L1272

LibVlcSharp.WPF.Sample/Controls.xaml.cs
```diff
@@ -0,0 +1,55 @@
...
+
+            InitializeComponent();
+            
+            PlayButton.Click += PlayButton_Click;
+            StopButton.Click += StopButton_Click;
+        }
+
+        private void StopButton_Click(object sender, RoutedEventArgs e)
+        {
+            parent.Dispatcher.Invoke(() => {
```
mfkl: off topic/review question: Do you think it'd make sense for the binding to offer a dispatcher callback registration, so users dont have to do just that on most calls?

LibVLCSharp.WPF/ForegroundWindow.xaml.cs
Ch4rg3r: Not sure when this actually can happen. Most likely would only happen, if no Window with Hwnd is provided, but then you wouldn't see anything anyway

LibVlcSharp.WPF.Sample/Controls.xaml.cs
Ch4rg3r: I'm not sure how to implement it inside the Control to work for all calls.

Ch4rg3r: seems like it's not needed anymore... just tested it without invoke (ThreadIDs are the same)

jeremyVignelles: @mfkl : I think it might be interesting to avoid a lot of questions like this one : https://github.com/ZeBobo5/Vlc.DotNet/issues/444

jeremyVignelles: Yes, it's not needed : You are in a UI event handler, which runs on the dispatcher.

Ch4rg3r:
I guess that's it so far. Should be completly fixed and working

jeremyVignelles:
Thanks for your reactivity.
You did a really good job, and thank you for that. You even managed to solve the Airspace issue in a clean manner, congratulations for that.

For me, the WPF control is OK. However, the sample looks more like a test project. I think that we can choose only one of the two Examples and make it clean and commented for future users. We can keep the structure with a "launcher" window, or make a "Minimal" sample, as made in Vlc.DotNet.

I just saw one bug : When the video is playing, closing the video window doesn't stop the player (audio is still playing). Please Dispose() properly the video control when the window is closing, so that users do the same. You might have to do that in another thread...

Ch4rg3r:
I will consider that...
Well the Airspace fix wasn't my idea, i jsut changed necessary parts.
I did the sample this way, to show 3 possible ways of using the control with child elements...

mfkl:
> Do you think MediaPlayer and LibVLC should use an instance, to prevent ObjectNullException??

I don't understand. What do you mean use an instance?

Ch4rg3r:
I mean the something like
```
get
{
    if(_mediaplayer==null)   
        _mediaplayer= new MediaPlayer()
    return _mediaplayer
}
```
Same for `LibVLC`

mfkl:
No, if you're using the provided `VideoView` control, you agree to have the `LibVLC` and `MediaPlayer` object's lifecycles bound to the control's view lifecycle. They won't be null ever (since newed in the ctor and disposed on dispose()) and I don't see a reason for your lazy loading suggestion. 
If a user needs more custom access, they have access to the classes and can do whatever they need.

jeremyVignelles:
I call that a lazy initialization (as in the Lazy<T> class). 
I think that errors might be hard to spot in those kind of patterns, because a call to a getter isn't meant to actually modify something. Then, later, you don't pay attention, call that from multiple threads and crash... then you lock everything and have perf penalties...

Ch4rg3r:
True, might be more often the case than NullException

mfkl:
Can you tick that checkbox so I can push on that PR please?
https://stackoverflow.com/questions/20928727/adding-commits-to-another-persons-pull-request-on-github

mfkl:
Made a PR on your fork. Please review the changes and next time use `git mv` to rename files ;-)

Ch4rg3r:
Merged.

mfkl:
https://twitter.com/martz2804/status/1019867718969782273

@Ch4rg3r thanks for your contribution, it's now live on nuget ;-) Feel free to keep those PRs coming!

Ch4rg3r:
Thanks for merging and for your work. Great library :thumbsup:

PR #2, opened 2018-07-22
GTK# / Xamarin.Forms.Gtk implementations

jeremyVignelles:
Here is the long-awaited GTK# pull request.

However this is still a WIP:

- [X] Write the code for the Gtk# implementation
- [X] Make it work on Windows
- [X] Make it work on Linux
- [X] Make it work on MacOS
- [X] Write the code for Xamarin.Forms using the GTK# backend
- [ ] Update LibVLC.Mac package to support net40/.NET Core projects https://github.com/mfkl/libvlc-nuget/pull/15
- [ ] Solve the initialization bug in Xamarin.Forms (discussion by mail with Xamarin.Forms.Gtk author)
- [ ] Check that the Forms.Gtk sample works on all three platforms
- [ ] Write installation instructions
- [ ] Write a linux-specific sample for Gtk# using snap ? (Requires videolan and gtk-sharp provide a snap packages...)

/cc @mfkl @Ch4rg3r 

PS: I rebased this branch on master and forced push. It seems that the comments on the commit were overwritten.


jbkempf:
Looks very cool!

Just a few questions:
- why not merge 2 and 4?
- why commiting the core-code and the sample in the same commit? I would split them,
- why run XInitThreads from libX11.so.6 and not libX11.so ?

jeremyVignelles:
Thanks j-b for your review,

I'm not familiar with rewriting history in multiple commits. I usually write commits in the order I write code and make fixes, and squash merge PR if I need to keep the history clean. If VideoLAN's policy is to keep code and samples separated, I'm fine with that.

Regarding ` libX11.so.6`, I agree, it should be just `libX11`

@jbkempf, do you have any opinions regarding the call to `XInitThreads`? Should it be always called in the GTK# binding, or should it be up to the caller to call that? Should we expose the `XInitThread` DllImport code, so that they don't have to write the NativeReference class?

Also, do you think it's worth writing a snap packaging sample? Will VideoLAN provide a "snap part" for libvlc?

jbkempf:
- git rebase -i origin/master is your best friend.
- for X11, I would always call it, yes.
- maybe, but we'll see that later.

jeremyVignelles:
Ok, I'll do that later.

Any reason why XInitThreads isn't called in vlc itself?
What will happen when GTK# will support GTK3 on wayland? :smile: (just kidding, we will need to make some edits in our code anyway)

jbkempf:
See vlc_xlib_init()

LibVLCSharp.Gtk.Sample/NativeReferences.cs
```diff
@@ -0,0 +1,20 @@
...
+        /// <summary>
+        /// Gets the window's XID
+        /// </summary>
+        /// <remarks>Linux X11 only</remarks>
+        /// <param name="gdkWindow">The pointer to the GdkWindow object</param>
+        /// <returns>The window's XID</returns>
+        [DllImport("libX11.so.6", CallingConvention = CallingConvention.Cdecl)]
+        public static extern int XInitThreads();
+    }
+}
```
mfkl: I believe you could put this in `Core.Initialize()`.

LibVLCSharp.Platforms.Gtk/NativeReferences.cs
```diff
@@ -0,0 +1,29 @@
...
+        /// <summary>
+        /// Gets the window's XID
+        /// </summary>
+        /// <remarks>Linux X11 only</remarks>
+        /// <param name="gdkWindow">The pointer to the GdkWindow object</param>
+        /// <returns>The window's XID</returns>
+        [DllImport("libgdk-x11-2.0.so.0", CallingConvention = CallingConvention.Cdecl)]
+        public static extern uint gdk_x11_drawable_get_xid(IntPtr gdkWindow);
+    }
+}
```
mfkl: These definitions belong to the `Internal` struct of the `MediaPlayer`.

jeremyVignelles:
j-b : I meant libvlc, sorry
mfkl : good points

jbkempf:
@jeremyVignelles well, because libVLC cannot know if you did it already or not, previously.

jeremyVignelles:
Then why should we make this assumption in LibVlcSharp? At first, I didn't want to take that decision for the developper because libvlc did not call it itself, and because I'd like to have an API that ressembles what users can find on the web for other languages so that they can easily translate code from one language to another.

LibVLCSharp.Gtk.Sample/Program.cs
```diff
@@ -0,0 +1,47 @@
...
+    using System;
+
+    public class Program
+    {
+        public static void Main()
+        {
+            // Initializes X threads before calling VLC. This is required for vlc plugins like the VDPAU hardware acceleration plugin.
+            if(RuntimeInformation.IsOSPlatform(OSPlatform.Linux))
+            {
+                NativeReferences.XInitThreads();
```
mfkl: `This function returns a nonzero status if initialization was successful; otherwise, it returns zero.`
Should we check for a return value here? And output a warning if that failed?

mfkl:
> Solve the initialization bug in Xamarin.Forms (discussion by mail with Xamarin.Forms.Gtk author)

@jsuarezruiz Hey Javier, could you please have a quick look at that Custom Renderer timing issue?

LibVLCSharp.Gtk.Sample/Program.cs
jeremyVignelles: Yes, we should

jsuarezruiz:
Yes, let me test and review everything :)
Looks nice!

jeremyVignelles:
> Why don't name it after bindings
Didn't understand that statement.

> My oppinion is, that GTK3 with glade is the better way, i got better results and it's easier for Cross-Platform development.
I agree it's easier to use and so on, that's why I suggested to implement this one as well. My problem is : we still need mono/gtk-sharp for XF.Gtk

jeremyVignelles:
mono/gtk-sharp plans on supporting GTK3 someday (there are already beta out there). Granted, nothing has changed recently, but when (if ever) it is out, it will be confusing to have GTK2 / GTK3

Ch4rg3r:
'If' ... then it's confusing, right now it's confusing if you need GTK2 and GTK3 and you don't know which one refers to GtkSharp for 3 or 2

jeremyVignelles:
We did some test yesterday to build that on macOS. Xamarin.Gtk built but failed to load some native libs at runtime. We still need some time to finish that, and then, I will then rebase and squash the commits. 

I can indeed make two separate PR, that would be nice to ship something.

@jsuarezruiz : Did you have time to look at that XF issue?

Yesterday, I rebased the branch, and felt like I lost some comment in the process. @Ch4rg3r, was that you that told me that we could simplify the references to gtk sharp?

Ch4rg3r:
References? I'm not sure, i justed mentioned the issue of versioning of GTK# (Version 2 & 3) related to Mono

jeremyVignelles:
I meant these lines : https://github.com/videolan/libvlcsharp/blob/f4dabcd702b8ae7e5fad9eec6f3cf2dc6f61eeeb/LibVLCSharp.Forms.Platforms.Gtk/LibVLCSharp.Forms.Platforms.Gtk.csproj#L14-L37
I'm pretty sure someone gave me a solution, and I can't remember...

Ch4rg3r:
I always had problems with GTK# on windows, but because i used GTK# 3. For 2 i could simply add references from Mono package

Ch4rg3r:
https://github.com/videolan/libvlcsharp/blob/f4dabcd702b8ae7e5fad9eec6f3cf2dc6f61eeeb/LibVLCSharp.Forms.Platforms.Gtk/LibVLCSharp.Forms.Platforms.Gtk.csproj#L14-L37

Is this necessary for GTK2? I thought to use it on Windows and Linux it requires Mono to be installed. Shouldn't the assemblies be availible in GAC and on Linux in fodlers with Path variable, so they are located and loaded automatically?

jeremyVignelles:
I just cleaned my inbox and found the post I was referring about references simplification : https://github.com/mono/gtk-sharp/issues/231#issuecomment-414152590

:tada:

jeremyVignelles:
Replaced by #20

PR #3, opened 2018-07-24
Fixed build warnings

jeremyVignelles:
Removed Fody on Android Sample project

(It's just because I don't like to see orange and red on my projects when building ^^ )

### DIFF
--- a/LibVLCSharp.Android.Sample/FodyWeavers.xml
+++ b/LibVLCSharp.Android.Sample/FodyWeavers.xml
@@ -1,3 +1,0 @@
-<?xml version="1.0" encoding="utf-8"?>
-<Weavers>
-</Weavers>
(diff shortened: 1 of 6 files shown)


mfkl:
Thanks for this.

I guess we might as well remove everything related to it:

https://github.com/videolan/libvlcsharp/blob/master/LibVLCSharp/Shared/ApiVersion.cs (we will have to bring it back for 4.0).
https://github.com/videolan/libvlcsharp/blob/e6694dda7b16850f8b8d5d6bc6c8b645e318f64a/LibVLCSharp.Tests/LibVLCSharp.Tests.csproj#L9
https://github.com/videolan/libvlcsharp/blob/master/LibVLCSharp/FodyWeavers.xml
https://github.com/videolan/libvlcsharp/blob/master/LibVLCSharp.Tests/VersionCheckTests.cs

I can't do much about the warnings from the automated generated C# code from Java interop.

mfkl:
Merged

PR #4, opened 2018-07-25
Corrects design mode

kakone:
Corrects design mode
Removes unnecessary things
Adds the possibility to update the content after control loading

### DIFF
--- a/LibVLCSharp.WPF/VideoView.cs
+++ b/LibVLCSharp.WPF/VideoView.cs
@@ -1,96 +1,114 @@
-﻿using System;
+﻿using LibVLCSharp.Shared;
+using System;
+using System.ComponentModel;
 using System.Windows;
 using System.Windows.Controls;
 
-using LibVLCSharp.Shared;
-
 namespace LibVLCSharp.WPF
 {
-    public class VideoView : UserControl, IVideoView, IDisposable
+    [TemplatePart(Name = PART_PlayerView, Type = typeof(System.Windows.Forms.Panel))]
+    public class VideoView : ContentControl, IVideoView, IDisposable
     {
-        double _controlWidth;
-        double _controlHeight;
-        readonly ForegroundWindow _foreground;
+        private const string PART_PlayerView = "PART_PlayerView";
+
+        static VideoView()
+        {
+            DefaultStyleKeyProperty.OverrideMetadata(typeof(VideoView), new FrameworkPropertyMetadata(typeof(VideoView)));
+        }
 
         public VideoView()
         {
-            var res = Application.LoadComponent(new Uri("/LibVLCSharp.WPF;component/Styles/VideoView.xaml", UriKind.RelativeOrAbsolute)) as ResourceDictionary;
+            var res = (ResourceDictionary)Application.LoadComponent(new Uri("/LibVLCSharp.WPF;component/Styles/VideoView.xaml", UriKind.RelativeOrAbsolute));
             Style = res["VideoViewStyle"] as Style;
 
-            _foreground = new ForegroundWindow(this);
+            if (!IsDesignMode)
+            {
+                ForegroundWindow = new ForegroundWindow(this);
 
-            Core.Initialize();
+                Core.Initialize();
 
-            LibVLC = new LibVLC();
-            MediaPlayer = new MediaPlayer(LibVLC);
+                LibVLC = new LibVLC();
+                MediaPlayer = new MediaPlayer(LibVLC);
 
-            SizeChanged += OnSizeChanged;
-
-            _controlHeight = Height;
-            _controlWidth = Width;
-
-            Loaded += VideoView_Loaded;
-            Unloaded += VideoView_Unloaded;
+                Unloaded += VideoView_Unloaded;
+            }
         }
 
-        void VideoView_Unloaded(object sender, RoutedEventArgs e)
+        ~VideoView()
         {
             Dispose();
         }
 
-        void VideoView_Loaded(object sender, RoutedEventArgs e)
-        {
-            MediaPlayer.Hwnd = ((System.Windows.Forms.Panel)Template.FindName("PART_PlayerView", this)).Handle;
+        private bool IsDesignMode => (bool)DesignerProperties.IsInDesignModeProperty.GetMetadata(typeof(DependencyObject)).DefaultValue;
+        private ForegroundWindow ForegroundWindow { get; }
+        private bool IsContentUpdating { get; set; }
 
-            if (Content != null)
+        public MediaPlayer MediaPlayer { get; private set; }
+        public LibVLC LibVLC { get; private set; }
+
+        private void VideoView_Unloaded(object sender, RoutedEventArgs e)
+        {
+            Dispose();
+        }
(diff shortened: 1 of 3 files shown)


kakone:
I also wanted to make it a true themable control but I need to have an AssemblyInfo file with a ThemeInfo attribute. I don't understand what kind of project is LibVLCSharp.WPF, it doesn't seem to be a classic WPF class library.

mfkl:
Thanks @kakone!

/cc @jeremyVignelles

mfkl:
> I don't understand what kind of project is LibVLCSharp.WPF, it doesn't seem to be a classic WPF class library.

I think it is, we just use `ExtrasEnableWpfProjectSetup` from https://github.com/onovotny/MSBuildSdkExtras which allows us to use the new sdk-style csproj, I believe.

jeremyVignelles:
@kakone : This is a WPF project that adopts "sdk-style" csproj, the new readable format used by .net core.

As it is currently unofficially supported (but will likely be when .NET core 3 with WPF will be out), we need to write some stuff ourself, or rely on a Sdk like MSBuild.Sdk.Extras, which does everything for us (XAML compilation...).

What's good with this new project format, is that you can edit it by hand, add your nuget dependencies there, put metadata here (Description...) and you don't need the attributes in AssemblyInfo.cs anymore (autogenerated in the build process, and will conflict with your definitions here).

It doesn't mean, however, that you can't put assembly-wide attributes. You can perfectly recreate an AssemblyInfo.cs or PutYourNameHere.cs and put your attributes there. (AssemblyInfo.cs is just a .cs file that gets built and integrated in the final assembly)

kakone:
Thanks, I added an AssemblyInfo file with the ThemeInfo attribute and it works great.

LibVLCSharp.WPF/VideoView.cs
```diff
@@ -1,96 +1,119 @@
...
+                MediaPlayer = null;
+            }
 
-        public MediaPlayer MediaPlayer { get; }
-        public LibVLC LibVLC { get; }
+            var libVLC = LibVLC;
+            if (libVLC != null)
+            {
+                libVLC.Dispose();
+                LibVLC = null;
```
mfkl: You're disposing `libVLC` and setting null to `LibVLC`. Is that on purpose? The small casing difference can lead to confusion.

LibVLCSharp.WPF/Themes/Generic.xaml
```diff
@@ -0,0 +1,26 @@
...
+    xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
+    xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
+    xmlns:wf="clr-namespace:System.Windows.Forms;assembly=System.Windows.Forms"
+    xmlns:local="clr-namespace:LibVLCSharp.WPF">
+
+    <Style TargetType="{x:Type local:VideoView}">
+        <Setter Property="Template">
+            <Setter.Value>
+                <ControlTemplate TargetType="{x:Type local:VideoView}">
+                    <Border Background="{TemplateBinding Background}"
```
mfkl: Any particular reason for why you added a `Border`?

LibVLCSharp.WPF/VideoView.cs
kakone: Yes, that's the purpose. I'm going to not use an intermediary variable, it's useless.
I added a finalizer that also call the Dispose method because in some cases (application shutdown for example), the Unloaded event is not called. So, in the other cases, the Dispose method will be called twice (in the Unloaded event and the finalizer).
I don't know if it is a good thing to call the Dispose method in the UnLoaded event because we can't remove the control from the visual tree without disposing the MediaPlayer.

LibVLCSharp.WPF/Themes/Generic.xaml
kakone: Yes, if we want that the BorderBrush, BorderThickness, ... properties of the Control works, we need to add a border in the ControlTemplate.

Ch4rg3r:
> it doesn't seem to be a classic WPF class library.

As they said it is intended. Also this is not a classic WPF Control either, as it wouldn't fix the airspace issue. Does the themable solution changes u did keep this working?

kakone:
I kept the ForegroundWindow class. The control is not fully themable due these airspace issues : you can't add some contents over the player in the theme file. But you can customize the rest (borders, ...).
Now, you can also update the Content property after the control initialization, and the Padding, BorderBrush, BorderThickness properties works well now.

Ch4rg3r:
I really don't get what's the point about this. Why it should be themable. You can place it everywhere, with every size and customize the sorrounding elements, and u can add themed child elements. The player itself doesn't need any theme, it's just a simple control, which adds the player. The player uses the full size of its control.

Ch4rg3r:
Performance. Maybe @mfkl can check the performance, if it slows down.

kakone:
I have no worries about performance, I'm pretty sure it will not change anything (on the contrary, it can only improve it, UserControl is more complex).

LibVLCSharp.WPF/Themes/Generic.xaml
mfkl: Ok. 

Can we make it hidden or transparent though? My only concern with this PR is that it makes the view look different than the views on other platforms (adds a border). I think we should have the same looking result everywhere for now. But making it themeable/updateable is totally valid!

jeremyVignelles: Just talked with @mfkl, and I quite agree with him : All control should have the same behavior as much as possible (to be embedded in a X.F app for example).
Adding a border isn't problematic if its BorderThickness is 0 by default, but on other platforms, we don't have borders on the control.
Is that an issue to let users put a Border around the control?

Ch4rg3r: @jeremyVignelles For plattform consistency it's obvisously different. In most cases you have to create"border" or any other visual yourself, because they miss the functionality of XAML. It makes it possible, to create a different visual appearance for WPF only. On the other hand, as long as it is possible to deactivate it, it doesn't matter. Also i don't think most platform implementations support any theming after all, except WPF.

kakone:
> Content in old case were UI Elements, which could be added and changed during runtime. This is what Example2 in WPF.Sample does!
No, the Content property is not changed in the Example2. A label is added to the existing StackPanel but you couldn't replace this StackPanel.

LibVLCSharp.WPF/Themes/Generic.xaml
kakone: >Adding a border isn't problematic if its BorderThickness is 0 by default
It's the case. By default, there is no border, no padding. I'm going to update the Example1 to remove the borders I added in order to test.

When you create a new control in WPF, there are these default properties (Padding, BorderBrush, ...) in the Control base class. In my opinion, when you create a custom control, these properties should work.

jeremyVignelles: @kakone : Fair enough, let's go for a border that is invisible by default.

PR #5, opened 2018-07-28
[Xamarin.Forms] Added support to Xamarin.Forms WPF Backend

jsuarezruiz:
There was enough prepared to have everything running very fast in the WPF backend of Xamarin.Forms.

imagen

- [x] Write the code for Xamarin.Forms using the WPF backend
- [x] Check that the Forms.WPF sample works
- [ ] Review feedback

/cc @mfkl @Ch4rg3r @jeremyVignelles


jeremyVignelles:
Forms.Gtk sample? what does Gtk have to do with Wpf?
Anyway, it's nice to have this implemented. I can't really review that now, but will try later

Ch4rg3r:
This is platform implemenation for Xamarin.Forms.WPF based on WPF-VideoView... Still GTK issues are there -.-

mfkl:
> Forms.Gtk sample? what does Gtk have to do with Wpf?
> This is platform implemenation for Xamarin.Forms.WPF based on WPF-VideoView... Still GTK issues are there -.-

There is no mention of GTK in this PR guys, this is about WPF and XF.

Thank you @jsuarezruiz, WPF backend for XF was on the roadmap as well. Will review soon!

jeremyVignelles:
There was a mistake in the PR description. It has been fixed 😉

With this implementation, will we have the airspace issue?

mfkl:
> There was a mistake in the PR description. It has been fixed 😉

Ok.

> With this implementation, will we have the airspace issue?

This is not a new implementation, just a custom renderer with our own `LibVLCSharp.WPF.VideoView` so same thing I'd say.

jeremyVignelles:
The Wpf implementation allows to add content "in" the control if I'm not mistaking. This can't be done in the current XF implementation, right?

jsuarezruiz:
Yes, there was an error in the description that I have changed to avoid confusion. This PR is for **Xamarin.Forms WPF** based on `LibVLCSharp.WPF.VideoView`.

As in Xamarin.Forms Android and iOS have access to [MediaPlayer](https://github.com/videolan/libvlcsharp/blob/f52df417e35fb0a083e235158c9ec75a3bbaba3a/LibVLCSharp/Shared/MediaPlayer.cs), with its properties and methods, do you mean that?

mfkl:
> do you mean that?

No I think he's talking about airspacing issue mentioned in https://github.com/videolan/libvlcsharp/pull/1#issuecomment-402395587

> This can't be done in the current XF implementation, right?

I'd say it is. Anything that can be done in pure WPF can be done in XF.WPF with custom renderers. Don't think that's any different here.

Ch4rg3r:
If i see it right, you can get the `Control` of `VideoViewRenderer` from code behind? Then you can add controls anyway.

jeremyVignelles:
You shouldn't have to access the platform specific VideoViewRenderer from the cross-platform view in Xamarin.Forms, I think it would go against the philosophy of it.

Ch4rg3r:
And not to the VideoView itself? Airspace issue can happen then...

jeremyVignelles:
You put the XF VideoView in your cross platform XF app, and the rendering is delegated to the platform-specific implementation, which, in that case, calls your Wpf VideoView.

In other words, your Airspace hack is hidden by a cross-platform abstraction, and we can't reach it

jeremyVignelles:
Hi @jsuarezruiz , I can't get it to build, I keep getting this message on build : 

> Xamarin.Forms tasks do not match targets. Please ensure that all projects reference the same version of Xamarin.Forms, and if the error persists, please restart the IDE.

Forms.Wpf.Sample is at 3.1.0.637273, but Forms.Sample is at 3.0.0.446417. Can you use the same version? Do you think we should upgrade X.F on master?

jeremyVignelles:
Thanks for your fix, I managed to build it.
Do you know why is there such a big window title bar on the sample?

mfkl:
> Do you know why is there such a big window title bar on the sample?

Likely because of https://github.com/mohachouch/WpfLightToolkit which is a dependency of Xamarin.Forms.Platform.WPF...

mfkl:
Do you know of a way to disable _all_ custom styling from the WPF Xamarin.Forms backend @jsuarezruiz ? Possibly in our custom renderer?

```
   protected override void Appearing()
        {
            base.Appearing();

            ((LightAppBar) Template.FindName("PART_TopAppBar", this)).Visibility = Visibility.Collapsed;
        }
```

Added this locally to your sample XF.WPF project. This is not really pretty but does the job. 

If you have a better way, I'm all ears. We also need to reset the blue theme from the top bar and buttons. I'd like to have a **default** view on all platforms. I also don't understand why the XF.WPF brings _custom_ styling enabled by default?

kakone:
> Do you know of a way to disable all custom styling from the WPF Xamarin.Forms backend @jsuarezruiz ? Possibly in our custom renderer?

It is the default style of XF WPF Window, why disable it in our custom renderer ?There must be a HasNavigationBar property on the FormsApplicationPage class, you can set it to Visibility.Collapsed.
Otherwise, you can set your own [style](https://github.com/xamarin/Xamarin.Forms/blob/master/Xamarin.Forms.Platform.WPF/Themes/FormsWindow.xaml), removing all you don't want.

kakone:
> What should we do then?
>- Leave as-is : Airspace in Wpf in XF is really a small fraction of the use cases

I don't think it's a small fraction. We want to be able to put some play and stop buttons on top of the video.
>- Use the same logic on all XF platforms : If you want to write something on top of the video, put it inside the XF control

It would be nice but I don't see how we can add a child content with a custom ViewRenderer.

Ch4rg3r:
> It would be nice but I don't see how we can add a child content with a custom ViewRenderer.

Isn't it possible using a content property like it's done in pure WPF? The WPF VideoView itself is using this too. In that case the content handling changed, to prevent airspace.

kakone:
> Isn't it possible using a content property like it's done in pure WPF ?
No because the View class doesn't have a Content property and even it was the case, it would contain some Xamarin Forms controls that you can't add in the native WPF VideoView.

jeremyVignelles:
Do you mean that you can't do Container controls with X.F View? like Border, ListBox ... ? Is there any other alternative base class?

mohachouch:
@mfkl Why do you want to delete base styles? The basic styles have been added because in WPF, there is no notion of ToolBar Item, TabbedPage, CarouselPage, MasterDetailPage.

In this case, it is an error. The bar is supposed to be unvisible when there is no title and no toolbaritem in the content page.

You can change the default color in your App.xaml (WPF) : 
https://github.com/xamarin/Xamarin.Forms/blob/master/Xamarin.Forms.Platform.WPF/WPFResources.xaml

kakone:
We can't do a container with a custom ViewRenderer. It's only for individual controls, but not for layout containers.
So, if we want to try to do the same kind of hack for the airspace issue, we must reimplement the [ForegroundWindow](https://github.com/videolan/libvlcsharp/blob/master/LibVLCSharp.WPF/ForegroundWindow.xaml.cs) in pure Xamarin Forms.

mfkl:
Thanks for chiming in.

> @mfkl Why do you want to delete base styles? 

Because those are _custom_ arbitrary styles. Why blue is the default and not green? :D I want my minimize/maximize/close buttons and command bar color to look the same they do on WPF and XF.WPF. Don't you think that makes sense?

kakone:
> About the VideoView, why not create a copy with the changes for foreground under LibVLCSharp.Forms.Platforms:WPF and use this in Renderer?

Like I said before, you can't manage a Content in the view renderer.

> Or we add this inside WPF.VideoView, this would require to have a check if it's plain WPF or XF adn then choose the right foreground.

We need the native WPF.VideoView to render the video in a WindowsFormsHost. So, we need two controls : the native WPF.VideoView and the Xamarin Forms VideoView with a ViewRenderer. But, I think that our Xamarin.Forms.WPF VideoView can inherit from the ContentView class and we can try to do a ForegroundWindow clone in pure Xamarin Forms.

kakone:
@jsuarezruiz Is there a reason to target .Net Framework 4.7 ? Could you target the minimum one (4.6.1) ?

mohachouch:
@kakone Xamarin Forms WPF target 4.6.1 framework

jsuarezruiz:
Fixed merge error with master branch and changed the .NET version of the projects related to Xamarin.Forms WPF from 4.7 to 4.6.1.

mfkl:
Merged since I am the only who seems to have an issue with the styling :D

On nuget soon.

Thanks Javier!

PR #7, opened 2018-07-31
Simplifies and improves the WPF VideoView

kakone:
Removes the OnContentChanged method
Adds an example for styling


kakone:
I always ask myself if it's a good thing to dispose LibVLC when the VideoView is removed from the visual tree. The developer could want to remove it from the visual tree and add it later, so the MediaPlayer should not be disposed, isn't it ?

mfkl:
> I always ask myself if it's a good thing to dispose LibVLC when the VideoView is removed from the visual tree.

Yeah, I've been wondering about this as well, as that's the current behavior on all platforms at the moment. Not sure what's the right way.
- Keep as is. Most users will probably just need a view to show some video and then move on with their lives. If they need something like video background playback (on mobile), they should just implement their own view/services.
- Offer a "keep in background" feature with view tombstoning on all platforms. Opt in or opt out? Is that something that should be supported by libvlcsharp?

kakone:
It joins this [problematic](https://code.videolan.org/videolan/LibVLCSharp/issues/13). Perhaps that we can dispose LibVLC only in the case we have done the instantiation. If the developer pass it own LibVLC/MediaPlayer, we don't dispose it.

mfkl:
This PR looks tightly coupled to the design discussions over at https://github.com/videolan/libvlcsharp/pull/8 like you mention, but also brings other improvements (samples + databinding). What should we do?

kakone:
You can merge it because it simplifies the control. I will remove the dispose method later with the changes of #8.

kakone:
I close this PR. I will look at this after the changes of #8.

PR #8, opened 2018-08-03
Adds a MediaSource class

kakone:
To illustrate the discussion https://code.videolan.org/videolan/LibVLCSharp/issues/13

I only tested quickly on Android / XF.Android and WPF


jeremyVignelles:
I don't really understand:

- How it works
- Which benefits does it bring
- Why is there a CustomMediaSource and a MediaSource?
- Why is there a SetWindowHandle in the CustomMediaSource? SetWindowHandle seems winforms specific to me

kakone:
> How it works
Which benefits does it bring

You just need to create a MediaSource (or a CustomMediaSource) and pass it to the VideoView. The main advantage is that you can declare the MediaSource in your ViewModel in bind it to your VideoView.Source property.
Other advantage : you can instantiate yourself only one LibVLC (in async way if you want) with multiple MediaPlayer.
Third advantage : I think it will easiest to do a background task with a MediaPlayer.

> Why is there a CustomMediaSource and a MediaSource?

The MediaSource is the more simple to use : the developer don't instantiate libVLC, nor the MediaPlayer. He can only pass an URI and play.
The CustomMediaSource is for advanced developers. They instantiate their libVLC and pass a MediaPlayer to the CustomMediaSource. They are responsible for disposing.

> Why is there a SetWindowHandle in the CustomMediaSource? SetWindowHandle seems winforms specific to me

I just remove it

kakone:
> I still can't understand the semantic behind the IMediaSource. If it is a "source", why does it have setters for Hwnd/AndroidContext...

I don't know how I can do otherwise. We need to set the MediaPlayer.Hwnd/AndroidContext property of the MediaPlayer class, that's why the ISource interface have these properties (the View need to pass this handle/android context, it doesn't need to know other thing).

> If I understand correctly, the only use of I(LibVlcMedia)Source is to wrap both LibVlc and a MediaPlayer.

That's right

> Why can't we pass a MediaPlayer directly?

You are right, I did a new commit so we can pass directly a MediaPlayer (so I removed the base class). The MediaPlayer implements now the ISource interface.

jeremyVignelles:
> The MediaPlayer implements now the ISource interface.
Why do we need the interface in the first place?
Do you plan to have other implementations of the media player?

kakone:
> Why do we need the interface in the first place?
Do you plan to have other implementations of the media player?

I don't plan to have other implementations of the media player, but other implementations of the ISource interface, yes. For example, in background task scenarios, I thought to have an ISource implementation that communicate between a MediaPlayer in background task and the VideoView.

LibVLCSharp/Shared/ISource.cs
```diff
@@ -0,0 +1,15 @@
+﻿using System;
+
+namespace LibVLCSharp.Shared
+{
+    public interface ISource
+    {
+#if WINDOWS
+            IntPtr Hwnd { set; }
+#elif ANDROID
+        void SetAndroidContext(IntPtr handle);
```
mfkl: Missing a tab indentation

LibVLCSharp/LibVLCSharp.csproj
```diff
@@ -41,5 +41,12 @@
   <ItemGroup Condition="$(TargetFramework.StartsWith('Xamarin.Mac'))">
     <Compile Include="Platforms\Mac\**\*.cs" />
   </ItemGroup>
+  <ItemGroup>
+    <None Remove="Shared\CustomMediaSource.cs" />
+    <None Remove="Shared\IMediaSource.cs" />
+    <None Remove="Shared\ISource.cs" />
+    <None Remove="Shared\MediaSource.cs" />
+    <None Remove="Shared\SourceChangedEventArgs.cs" />
+  </ItemGroup>
```
mfkl: You can remove those, they're automatically included.

LibVLCSharp/Shared/MediaSource.cs
```diff
@@ -0,0 +1,82 @@
+﻿using System;
+using System.Threading.Tasks;
+
+namespace LibVLCSharp.Shared
+{
+    public class MediaSource : IMediaSource
+    {
+        static MediaSource()
```
mfkl: Maybe the constructor is worth a warning as comments that it initializes libvlc (and makes native calls). That's the only object from the wrapper that doesn't exist in the C API and makes native calls.

LibVLCSharp/Shared/MediaSource.cs
```diff
@@ -0,0 +1,82 @@
...
+            set { MediaPlayer.NsObject = value; }
+        }
+#endif
+
+        public static Task<MediaSource> CreateAsync(params string[] cliOptions)
+        {
+            return Task.Run(() => new MediaSource(cliOptions));
+        }
+
+        public static Task<MediaSource> CreateFromUriAsync(string uri)
```
mfkl: Maybe add an extra ctor with both `cliOptions` and `uri` for lazy devs? :-P

LibVLCSharp/Platforms/Android/VideoView.cs
```diff
@@ -8,60 +8,77 @@
 using LibVLCSharp.Shared;
 
 using Org.Videolan.Libvlc;
+using static Org.Videolan.Libvlc.AWindow;
```
mfkl: I guess that's necessary now

LibVLCSharp/Shared/ISource.cs
```diff
@@ -0,0 +1,15 @@
+﻿using System;
+
+namespace LibVLCSharp.Shared
+{
+    public interface ISource
```
mfkl: Are you sure both `IMediaSource` and `ISource` are needed? Wouldn't one interface be enough?

LibVLCSharp/Shared/MediaSource.cs
kakone: I'm going to update the `CreateFromUriAsync(string uri)` to `CreateFromUriAsync(string uri, param string[] cliOptions)`

I removed the signature `MediaSource.CreateFromMediaPlayer(MediaPlayer mediaPlayer)` because the MediaPlayer implements ISource. So, you can pass a MediaSource class or a MediaPlayer class to your VideoView. I don't know what is best :
- let as this, or
- remove the ISource interface from MediaSource (and the developer should pass MediaSource.MediaPlayer), or
- remove the ISource interface from the MediaPlayer and encapsulate the MediaPlayer to a class that only implements the ISource interface (in this case, we add the signature MediaSource.CreateFromMediaPlayer)

LibVLCSharp/Shared/ISource.cs
kakone: You are right but we only need the ISource interface (I have background task scenario in mind, I will try to do an example, and in this case, we don't want to have a MediaPlayer in the interface).

LibVLCSharp/Shared/MediaSource.cs
```diff
@@ -0,0 +1,137 @@
...
+
+        /// <summary>
+        /// Finalizer
+        /// </summary>
+        ~MediaSource()
+        {
+            Dispose();
+        }
+
+        private static Lazy<bool> CoreInitialized { get; } = new Lazy<bool>(() =>
```
mfkl: I believe this property belongs in `Core.cs` (and it could be called multiple times, which is a case not currently handled right now but not in the scope of this PR, which is already quite big). Not quite sure about your `Lazy` choice there though.

LibVLCSharp/Shared/MediaSource.cs
```diff
@@ -0,0 +1,137 @@
...
+        public void SetAndroidContext(IntPtr aWindow)
+        {
+            MediaPlayer.SetAndroidContext(aWindow);
+        }
+#elif COCOA
+        public IntPtr NsObject
+        {
+            set { MediaPlayer.NsObject = value; }
+        }
+#endif
```
mfkl: @jeremyVignelles mentioned that we don't really need those helpers as the properties are public and already `#if def'ed`. I agree with him.

LibVLCSharp/Shared/MediaSource.cs
kakone: You are totally right. I'm going to move it to `Core.cs`. I will not need such a property in `Core` class, I can use the static constructor.

jeremyVignelles: A static constructor doesn't feel sane to me. Can you use a `IsInitialized` field, paired with the `Core.Initialize()` method, that will check that the core is not already initialized before doing anything else?

jeremyVignelles:
The only added-value of the `MediaSource` class is that it creates and dispose `LibVlc` and `MediaPlayer` at once. I'm still wondering if it's worth it or not...

We could probably data-bind on the `MediaPlayer`.

LibVLCSharp/Shared/MediaSource.cs
kakone: I need these properties because it's the implementation of the ISource interface. But, I'm going to remove the `public` keyword (replacing by the explicit interface implementation `IntPtr ISource.Hwnd`, ...).

kakone:
> We could probably data-bind on the `MediaPlayer`

Currently, you can databind directly on the `MediaPlayer` (because the `MediaPlayer` implements the `ISource` interface). The `MediaSource` is just an helper class but I think we should keep it, I find that the CreateAsync method is useful.

LibVLCSharp/Shared/MediaSource.cs
kakone: I do not see the problem with a static constructor but I can add a thread safe IsInitialized property.
In my mind, I was going to do a static constructor that do the initialization and the Core.Initialize method was going to be empty (just to call the static constructor). It's just an implementation problem but the result is the same : we have to be able to call Core.Initialize several times and do nothing if the initialization is already done.

kakone:
Good understanding ;)

> The naming of MediaSource is still confusing to me
Yes, I think we must clearly name it with Helper suffix. MediaHelper ?

> I'm also wondering like @jeremyVignelles said if we can't get away with just a MediaPlayer property on the view
I also think it will be clearer for the developer. But, it will complicate the background playback scenario. I'm going to do do some changes. I will add a MediaPlayer property and I will keep the access to a Source property internally with an interface for the playback scenario.

mfkl:
Closing for now as part of this PR has been merged with https://github.com/videolan/libvlcsharp/pull/18

Feel to make a new PR to create a new sample with more advanced components contained in this PR. And thanks for your comments, work and feedback on the API @kakone

PR #9, opened 2018-08-06
Simplify event handling + fix iOS

mfkl:
- Simplify events. No more event managers, events belong to the objects (aligns with standard .NET API designs).
- Fixes https://code.videolan.org/videolan/LibVLCSharp/issues/15, iOS reverse callbacks. 
https://developer.xamarin.com/api/type/ObjCRuntime.MonoPInvokeCallbackAttribute/

WIP.

- [ ] events are done but more callbacks remain (look for "UnmanagedFunctionPointer").

 - [ ] make sure to squash and cherry pick things correctly (first 2 commits are unrelated).

I do NOT like riddling the code with `#if #else` directives but not sure how to do it more cleanly.


mfkl:
Currently looking at how to do this with Fody. Here's the list of changes to apply to the AST.

- Add a `using ObjCRuntime;`in relevant classes,
- Set all `EventHandler` fields to `static`,
- Set all relevant methods (tagged with `[AOT]` for example) as `static`. 
- Add`[MonoPInvokeCallback(typeof(type-of-the-correct-delegate))]` to all relevant methods. 

Tricky part is to make this run after IL but before AOT (iOS only).

Other ways to do this not using fody:
- MSBuild task https://natemcmaster.com/blog/2017/07/05/msbuild-task-in-nuget/
- Partial classes (I don't like)
- Platform specific implementation hidden behind interface
- ??

Insights welcome.

jeremyVignelles:
> - MSBuild task https://natemcmaster.com/blog/2017/07/05/msbuild-task-in-nuget/
> - Partial classes (I don't like)

As I thought it, the MSBuild task would generate code (several .cs file) as partial classes that would handle all the targetPlatform-specific code inside the same class.

mfkl:
the fody solution wasn't right, went with a `#if IOS` solution for now. Shipped in 0.0.1

PR #11, opened 2018-09-10
Initial opengl callback interop

mfkl:
This adds new interop callbacks for opengl rendering. It will allow integration with game engines and other graphics libraries such as https://github.com/mellinoe/veldrid.

For now, it's opengl/opengles only (no direct3D yet).

I'd like for a sample to be added to this PR before merging.

Nightly or libvlc 4.0 only.

### DIFF
--- a/LibVLCSharp/Shared/MediaPlayer.cs
+++ b/LibVLCSharp/Shared/MediaPlayer.cs
@@ -629,6 +629,14 @@ namespace LibVLCSharp.Shared
                 EntryPoint = "libvlc_media_player_set_android_context")]
             internal static extern void LibVLCMediaPlayerSetAndroidContext(IntPtr mediaPlayer, IntPtr aWindow);
 #endif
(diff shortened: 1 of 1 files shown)


mfkl:
Waiting for libvlc API to stabilize to finish this.

Ch4rg3r:
Couldn't we make the OpenGL implementation (and Veldrid) availible by using the Veldrid library and their SDL2/OpenGL implementation? They have native support for all necessary classes/functions. 
This wouldn'n be too difficult.

mfkl:
> Couldn't we make the OpenGL implementation (and Veldrid) availible by using the Veldrid library and their SDL2/OpenGL implementation?

No, we want libvlc to do the decoding and provide the decoded textures (which could be sent after that to veldrid for display, sure). Still waiting for that API to stabilize.

mfkl:
Closing that for now.

PR #12, opened 2018-09-11
Initial Unity.Android setup

mfkl:
This will be deployed on the Unity Store alongside an (opensource) unity plugin.


mfkl:
Will merge soon unless you have objections @jeremyVignelles

PR #14, opened 2018-09-19
support netstandard1.1 and net40

mfkl:
- [x] add support for netstandard1.1 target

- [x] add support for net40 target

- [ ] review


LibVLCSharp/Shared/LibVLC.cs
```diff
@@ -596,9 +573,11 @@ public MediaDiscoverer.Description[] MediaDiscoverers(MediaDiscoverer.Category c
         {
             var arrayResultPtr = IntPtr.Zero;
             var count = Native.LibVLCMediaDiscovererListGet(NativeReference, category, ref arrayResultPtr);
-
+#if NETSTANDARD1_1
+            if (count == 0) return new MediaDiscoverer.Description[0];
+#else
             if (count == 0) return Array.Empty<MediaDiscoverer.Description>();
-
+#endif
```
jeremyVignelles: Below, you replaced with `if (count == 0) return new RendererDescription[0];` for all frameworks, which is the most compatible.
Can you use the same fix for both?

LibVLCSharp/Shared/LibVLC.cs
```diff
@@ -667,8 +646,8 @@ public MediaDiscoverer.Description[] MediaDiscoverers(MediaDiscoverer.Category c
                     updateProgress(dlg, position, text);
                 }
             };
-
-            _dialogCbsPtr = Marshal.AllocHGlobal(Marshal.SizeOf<DialogCallbacks>());
+
```
jeremyVignelles: Remove whitespace

LibVLCSharp/Shared/LibVLC.cs
jeremyVignelles: I found some articles saying Array.Empty caches the new T[0] call for future use : http://justinvp.com/2015/07/20/array-empty/ 
You can keep that for perf if you want, but chose only one solution.

LibVLCSharp/Shared/Utf8StringMarshaler.cs
```diff
@@ -9,10 +9,11 @@ namespace LibVLCSharp.Shared
     /// <summary>
     /// Marshal unicode string param to utf-8 string,usage:[MarshalAs(UnmanagedType.CustomMarshaler, MarshalTypeRef = typeof(Utf8StringMarshaler))]
     /// </summary>
+#if !NETSTANDARD1_1
```
jeremyVignelles: Put this line above comment

jeremyVignelles:
What is the work needed to bring net40 once netstandard 1.1 is done ? Is there anything else to do than adding the target in the csproj?

mfkl:
> What is the work needed to bring net40 once netstandard 1.1 is done ? Is there anything else to do than adding the target in the csproj?

A few things but no deal breaker.

```
The name 'RuntimeInformation' does not exist in the current context
The name 'OSPlatform' does not exist in the current context
'Array' does not contain a definition for 'Empty'
The non-generic method 'Marshal.PtrToStructure(IntPtr, object)' cannot be used with type arguments
'Array' does not contain a definition for 'Empty'
'Task' does not contain a definition for 'Run'
```

LibVLCSharp/Shared/LibVLC.cs
mfkl: > chose only one solution.

If I want to keep using `Array.Empty` where possible, I can't choose "only one solution". 
It's either `new [0]` or a mix of both depending on what is available.

jeremyVignelles: OK, but then you could use the same #ifdef logic everywhere

mfkl:
> Is ad6b75e related to this PR?

No I will cherry pick it

LibVLCSharp/Shared/Utf8StringMarshaler.cs
mfkl: Why? Then the comment is not included and I have to have it twice

jeremyVignelles: You don't have a #else, so in case you're in netstandard, you have a comment on nothing

mfkl: > You don't have a #else

Yes, I do :-P It's just below, I think you missed it?

https://github.com/videolan/libvlcsharp/commit/2c63039b71c6e378e31d9a1b173dc599900a0d33#diff-60a9527f7fdf98b6c531858aea34e879

Will merge today unless you have more remarks

mfkl:
Should `NsObject` and `XWindow` be exposed for the `net40` target?

Assuming mono can run a `net40` dll, I guess it should.

jeremyVignelles:
this needs to be tested , but yes I think it should. What difference do you make between Desktop and Modern desktop?

mfkl:
> What difference do you make between Desktop and Modern desktop?

Writing `#if DESKTOP || NET40` was confusing me because `net40` is essentially running only on desktop, 

Those 2 defines cannot be merged though because `net40` does not support `RuntimeInformation.IsOSPlatform` (because it's old). Also, `net40` could run on XP according to Wikipedia ".NET Framework 4.0 is supported on Windows XP (with Service Pack 3)".

Opposite of that is modern I guess so I renamed it.

mfkl:
I'll check net40 soon on mac with mono. 

We have 2 problems with the netstandard1.1 build though, check my last commit.

```
typeof(LibVLC).Assembly;
File.Exists(libraryPath)
```
These APIs are not in netstandard1.1. Only realizing now because of the compilation directives. Looking for alternatives.

LibVLCSharp/LibVLCSharp.csproj
```diff
@@ -22,7 +22,9 @@
     <DefineConstants>$(DefineConstants);UNITY_ANDROID</DefineConstants>
   </PropertyGroup>
   <ItemGroup>
-    <PackageReference Include="System.Runtime.InteropServices.RuntimeInformation" Version="4.3.0" />
+    <PackageReference Condition="'$(TargetFramework)' != 'net40'"
```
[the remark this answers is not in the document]
mfkl: I'll have to check (but I recall it's needed somewhere)

LibVLCSharp/Shared/Core.cs
```diff
@@ -79,15 +92,41 @@ static IntPtr PreloadNativeLibrary(string nativeLibrariesPath, string libraryNam
...
             return Native.LoadLibrary(libraryPath);// TODO: cross-platform load
         }
+
+        static bool IsWindows
+        {
+            get
+            {
+#if NET40
+            return Environment.OSVersion.Platform != PlatformID.MacOSX 
+                && Environment.OSVersion.Platform != PlatformID.Unix;
```
jeremyVignelles: Isn't `Win32NT` enough here? Do you count Xbox as "Windows" ? Do we really need to care about win95 and win98?

jeremyVignelles: Source : https://docs.microsoft.com/en-us/dotnet/api/system.platformid?view=netframework-4.7.2

LibVLCSharp/Shared/Core.cs
```diff
@@ -111,6 +150,10 @@ internal static class Constants
         /// The name of the folder that contains the per-architecture folders
         /// </summary>
         internal const string LibrariesRepositoryFolderName = "libvlc";
+
+        internal const string Windows = "msvcrt";
```
jeremyVignelles: I don't think that those 3 constants have a good naming. Either you can create a static class named `StandardLibraryPath` and move them there, or you will need to prefix those constants

LibVLCSharp/Shared/MediaPlayer.cs
```diff
@@ -10,621 +10,618 @@ public class MediaPlayer : Internal
...
             internal static extern void LibVLCMediaPlayerSetXwindow(IntPtr mediaPlayer, uint drawable);
 
-            [SuppressUnmanagedCodeSecurity]
             [DllImport(Constants.LibraryName, CallingConvention = CallingConvention.Cdecl,
                 EntryPoint = "libvlc_media_player_get_xwindow")]
             internal static extern uint LibVLCMediaPlayerGetXwindow(IntPtr mediaPlayer);
-#if DESKTOP
-            [SuppressUnmanagedCodeSecurity]
+
+#if NETSTANDARD2_0 || NET40
```
jeremyVignelles: NETSTANDARD1_1 can't do that?

LibVLCSharp/Shared/RendererDiscoverer.cs
```diff
@@ -76,36 +76,34 @@ public class RendererItem : Internal
     {
         const int VideoRenderer = 0x0002;
         const int AudioRenderer = 0x0001;
+#if NETSTANDARD1_1
+        readonly Utf8StringMarshaler _utf8Marshaler = Utf8StringMarshaler.GetInstance();
```
jeremyVignelles: Can we just change the return type for all platforms to avoid the if?

mfkl: Not if we plan to use it as part of `DllImport` statements in which case it needs to be `ICustomMarshaler`

LibVLCSharp/Shared/MediaPlayer.cs
mfkl: if by _that_ you mean `SuppressUnmanagedCodeSecurity`, nope.

LibVLCSharp/Shared/RendererDiscoverer.cs
jeremyVignelles: `Utf8StringMarshaler` is a `ICustomMarshaler` anyway. You will only need a cast.

PR #17, opened 2018-10-18
initial versioning strategy doc

mfkl:
Feel free to contribute your opinions.

More info: https://code.videolan.org/videolan/LibVLCSharp/issues/22

PR #18, opened 2018-10-18
Support background playback by uncoupling the VideoView lifecycle from the Mediaplayer lifecycle

mfkl:
Following up on https://github.com/videolan/libvlcsharp/pull/8 

**This change breaks the public API**

Currently the VideoView is exposing a `LibVLC` and `MediaPlayer` which it instanciates when the view is created and dispose of when the view is disposed.

This change only exposes the `MediaPlayer`, which must be provided to the View, and the View does not dispose it when itself gets disposed. Allowing for background playback and such. It also provides a `MediaPlayerChanged` event.

- Migrate all samples to the new API
- Update XF
- WPF.Forms now targets .NET 4.7 (see last commit for info)

What additional APIs do we want on our VideoView (platforms and Forms)?

Do we want to bridge the `MediaPlayer` events directly onto the View?

Ideas:
https://github.com/martijn00/XamarinMediaManager/blob/develop/MediaManager.Forms/VideoView.cs
https://github.com/xamarin/xamarin-forms-samples/blob/master/CustomRenderers/VideoPlayerDemos/VideoPlayerDemos/VideoPlayerDemos/FormsVideoLibrary/VideoPlayer.cs

Since this breaks the public API, and this binding targets version 3 of libvlc, I think we should move the versioning from 0.0.3 to 3.0.0 or something. https://github.com/videolan/libvlcsharp/pull/17


jeremyVignelles:
Bridging the events on the view is really hard to maintain. In Vlc.DotNet, APIs are bridged only to the Forms control. Each time a new API is added (many are missing there), we always forget to edit the WinForms control.

However, for XAML developpers, having dependency properties is a long-asked feature.

Can we do something without having to add all APIs in all controls?

For the breaking change, can we go for 0.1.0 instead? Is the API stable yet?

mfkl:
> For the breaking change, can we go for 0.1.0 instead? Is the API stable yet?

Sure.

> Can we do something without having to add all APIs in all controls?

If the intended result is something like
```
<forms:VideoView x:Name="videoView" 
                Grid.Row="0"
                Grid.RowSpan="1"
                HorizontalOptions="FillAndExpand"
                VerticalOptions="FillAndExpand"
                AspectMode="AspectFill"
                AutoPlay="True"
                Media="{Binding ViewModel.Media}"
                MediaChanged="{Binding ViewModel.MediaChanged}"
                Buffering="{Binding ViewModel.Buffering}"
                Forward="{Binding ViewModel.Forward}"
                Backward="{Binding ViewModel.Backward}"
                PositionChanged="{Binding ViewModel.PositionChanged}"
                SeekableChanged="{Binding ViewModel.SeekableChanged}"
                VolumeChanged="{Binding ViewModel.VolumeChanged}"/>
```

Then I'm afraid we will have to bind them all.

Otherwise, we tell XAML developers that this approach is NOT supported and that they should use C# code. Do you see another way? Or are you thinking of another API?

jeremyVignelles:
{Binding} to events is not something that is supported in XAML(at least for WPF) either.

I think that most of the operations should happen in the ViewModel, where the MediaPlayer is stored. It gives you the best flexibility.

IMHO, the view is only there for rendering, as a computer screen is there only to display your computer's output.

My point is : the MediaPlayer is the ViewModel for the View, you can attach events in your view model, and attach it to the view:

```
<vlc:VideoView MediaPlayer="{Binding MediaPlayer}" />
```

jeremyVignelles:
Then, we need to discuss how we could support the seekbar:

```
<WhichControlIsTheSeekBar Position="{MediaPlayer.Position}" />
```

Should the MediaPlayer implement INPC? Should we provide an intermediate layer for binding?

kakone:
> {Binding} to events is not something that is supported in XAML(at least for WPF) either.

You can use an EventToCommand behavior to bind to events.

But I don't think either it's a good idea to expose all the events in the VideoView. Finally, in #8 (I didn't have time to finish it, sorry), I had the same idea, I wanted to have only one bindable property in the VideoView : MediaPlayer. The VideoView should stay minimalist. If you want a SeekBar, you can create a control for the seekbar with the same principle : only a MediaPlayer bindable property. In this case, everything is well disconnected. There is no point in having bindable properties for each event in the views when you have the MediaPlayer in your ViewModel.

Ch4rg3r:
Hi, i guess binding for MVVM should only have access to the mediaplayer, so developers can handle all the stuff they need themselfes. API on videoview requires too many changes and on API changings everything breaks

mfkl:
So I cleaned up my commits. Tested on iOS, Android and WPF. Remaining points are the following:
- There is a weird display bug on Android introduced by some code in this branch. See what I mean here https://imgur.com/a/yBoWyRT. Rotating the screen makes the display whole again.
- WPF: Very rarely, some timing issue will make it so that the HWND is not set on time and when VLC notices that no HWND has been set, it just creates a new one... Very hard to reproduce, and not specific to this code. Just a remark since I experienced it (again) while working on this.
- The `MediaPlayerChangedEventArgs`. Do we want it in `LibVLCSharp` for platform specific `VideoView`s or just in `LibVLCSharp.Forms`? I think we want it in `LibVLCSharp` as well, opinions welcome.
- I just added a very small VM to the Forms sample project for showing the databinding. I did not re-use all of the components introduce in the Forms Sample of https://github.com/videolan/libvlcsharp/pull/8. It's worth doing, but a more appropriate location for these would be https://code.videolan.org/mfkl/libvlcsharp-samples (trying to keep the samples in the libvlcsharp repo _minimal_ and focused)

mfkl:
OK, added a `DependencyProperty for WPF`

The `LibVLCSharp.WPF.Sample` works fine but the `LibVLCSharp.Forms.Sample.WPF` does not (opening a second window, MediaPlayer is set too late). Did not figure this out yet.

mfkl:
https://github.com/xamarin/Xamarin.Forms/pull/3482 of interest related to this

jeremyVignelles:
Shall LibVlcSharp.Forms be a drop-in replacement for the X.F implementation?

mfkl:
> Shall LibVlcSharp.Forms be a drop-in replacement for the X.F implementation?

If by that you mean that our APIs should (almost) exactly match theirs, I think _no_. Appealing idea in theory for developers, but I don't think that's realistic considering the different capabilities involved.

Also, 
> I don't think either it's a good idea to expose all the events in the VideoView.

That's the way they've gone it seems https://github.com/xamarin/Xamarin.Forms/blob/d2b56085017ed6c0ac414c5af501a453daa897a5/Xamarin.Forms.Core/MediaElement.cs

```
[RenderWith(typeof(_MediaElementRenderer))]
	public sealed class MediaElement : View, IMediaElementController
	{
		public static readonly BindableProperty AspectProperty =
		  BindableProperty.Create(nameof(Aspect), typeof(Aspect), typeof(MediaElement), Aspect.AspectFit);
		public static readonly BindableProperty AutoPlayProperty =
		  BindableProperty.Create(nameof(AutoPlay), typeof(bool), typeof(MediaElement), true);
		public static readonly BindableProperty BufferingProgressProperty =
		  BindableProperty.Create(nameof(BufferingProgress), typeof(double), typeof(MediaElement), 0.0);
		public static readonly BindableProperty CurrentStateProperty =
		  BindableProperty.Create(nameof(CurrentState), typeof(MediaElementState), typeof(MediaElement), MediaElementState.Closed);
		public static readonly BindableProperty DurationProperty =
		  BindableProperty.Create(nameof(Duration), typeof(TimeSpan?), typeof(MediaElement), null);
		public static readonly BindableProperty IsLoopingProperty =
		  BindableProperty.Create(nameof(IsLooping), typeof(bool), typeof(MediaElement), false);
...
```

mfkl:
> There is a weird display bug on Android introduced by some code in this branch. See what I mean here https://imgur.com/a/yBoWyRT. Rotating the screen makes the display whole again.

This is still an issue

mfkl:
Ok that last commit fixes it. Review and testing welcome people, I intend to merge this and deploy ASAP

PR #23, opened 2018-12-06
Set up CI with Azure Pipelines

mfkl:


mfkl:
Hi @kaylangan,

Thanks for reaching out.

So I'm indeed trying to setup CI with Azure DevOps for LibVLCSharp. However while this github repo welcomes PRs, this is a mirror repository and most of our development happens on our self hosted GitLab instance https://code.videolan.org/videolan/LibVLCSharp. My needs are:
- I'd like for Azure DevOps to build both GitLab merge requests and GitHub pull requests (in addition of regular CI builds).
- I'd rather not give Azure DevOps any VideoLAN GitHub rights, but read only would be OK I guess.
- I want to use YAML configuration instead of the visual designer (which doesn't seem possible with GitLab?).

Any help is welcome 👍

kaylangan:
@mfkl Sorry for the delayed reply.

What I suggest is that you set up two build definitions, one for GitHub and one for GitLab. For the GitHub build, you can use YAML and set it up using OAuth (as opposed to the GitHub app) so you don't have to give the Azure DevOps app any rights. For the GitLab build, you'll need to [use the "External Git" (generic) source](https://docs.microsoft.com/en-us/azure/devops/pipelines/repos/?view=vsts#supported-repository-types). Unfortunately, you are correct, we do not yet have YAML support for generic Git sources so you'd have to use the visual designer for now.

mfkl:
>  For the GitLab build, you'll need to use the "External Git" (generic) source.

OK, [I've done that](https://videolan.visualstudio.com/LibVLCSharp/_build/results?buildId=371&_a=summary). Is there any way to build Gitlab Merge requests from Azure DevOps? I don't a way to feed Azure DevOps with an external token to hook from the UI.

> you can use YAML and set it up using OAuth (as opposed to the GitHub app) so you don't have to give the Azure DevOps app any rights.

That still won't work.


When creating a new github service connection, I _did_ include `Recommended scopes -- repo, read:user, user:email, admin:repo_hook` in my token config in github.

PR #27, opened 2019-01-09
Readonly structs

mfkl:
pretty big PR with various fixes
- use readonly structs everywhere
- Add more comments
- fix most unit tests. running them from the VS runner is green with some logs 
```
[1/9/2019 3:39:15 PM Error] The active test run was aborted. Reason: Unhandled Exception: System.NullReferenceException: Object reference not set to an instance of an object.
Unhandled Exception: System.NullReferenceException: Object reference not set to an instance of an object.
```
Still green though. I suspect it has to do with process configuration with nunit. `new LibVLC()` starts a vlc qt app... Not sure how to configure this yet. Failing when ran from `nunit3-console.exe`.
- fix [OUT] marshalling for a few APIs
- refactor more marshaling logic to reduce code duplication
- remove unused code,
- remove unused dependencies on test project.

Considering all the changes and moving around of stuff, it probably breaks API (with namespace and all). But trivial to fix. Should probably bump to 0.2.0.

### DIFF
--- a/LibVLCSharp/Shared/Media.cs
+++ b/LibVLCSharp/Shared/Media.cs
@@ -548,7 +548,7 @@ namespace LibVLCSharp.Shared
         /// </remarks>
         public void ParseStop() => Native.LibVLCMediaParseStop(NativeReference);
 
-        /// <summary>Get media descriptor's elementary streams description</summary>
+        /// <summary>Get media descriptor's elementary streams description
         /// <para>address to store an allocated array of Elementary Streams</para>
         /// <para>descriptions (must be freed with libvlc_media_tracks_release</para>
         /// <para>by the caller) [OUT]</para>
@@ -559,6 +559,7 @@ namespace LibVLCSharp.Shared
         /// <para>Not doing this will result in an empty array.</para>
         /// <para>LibVLC 2.1.0 and later.</para>
         /// </remarks>
+        /// </summary>
         public MediaTrack[] Tracks => MarshalUtils.Retrieve(NativeReference, (IntPtr nativeRef, out IntPtr array) => Native.LibVLCMediaTracksGet(nativeRef, out array),
             MarshalUtils.PtrToStructure<MediaTrackStructure>,
             m => m.Build(),


LibVLCSharp/Shared/Helpers/MarshalUtils.cs
```diff
@@ -98,12 +94,119 @@ internal struct Native
             return resultList.ToArray();
         }
 
+        internal static TU[] Retrieve<T, TU>(IntPtr nativeRef, ArrayOut getRef, Func<IntPtr, T> retrieve,
```
jeremyVignelles: Could you please add XML comments on these complex methods: What are each for, how to use them... I'm not sure I'll remember after a while.

LibVLCSharp/Shared/Helpers/MarshalUtils.cs
```diff
@@ -98,12 +94,119 @@ internal struct Native
...
+
+            for (var i = 0; i < count; i++)
+            {
+                var ptr = Marshal.ReadIntPtr(arrayPtr, i * IntPtr.Size);
+                structure = retrieve(ptr);
+                var managedStruct = create(structure);
+                resultList.Add(managedStruct);
+            }
+
+            releaseRef(arrayPtr, count);
```
jeremyVignelles: try/finally to ensure it is always released?

LibVLCSharp/Shared/MediaPlayer.cs
```diff
@@ -8,7 +8,7 @@ namespace LibVLCSharp.Shared
 {
     public class MediaPlayer : Internal
     {
-        struct Native
+        readonly struct Native
```
jeremyVignelles: Why are all "Native" `readonly struct`? Wouldn't it make sense to have them `internal static class` instead?

LibVLCSharp/Shared/MediaPlayer.cs
```diff
@@ -1361,21 +1335,26 @@ public string AspectRatio
...
-        public bool SetSpu(int spu)
-        {
-            return Native.LibVLCVideoSetSpu(NativeReference, spu) == 0;
-        }
+        /// <summary>
+        /// Set Spu (subtitle)
+        /// </summary>
+        /// <param name="spu">Video subtitle track to select (id from track description)</param>
+        /// <returns>true on success, false otherwise</returns>
+        public bool SetSpu(int spu) => Native.LibVLCVideoSetSpu(NativeReference, spu) == 0;
```
jeremyVignelles: Why do you have a property with only a getter, and a `Set...()` method? I'd prefer to have `Get...()` methods too for symmetry (and also, I'm not really a fan of getters that allocates memory).

LibVLCSharp/Shared/MediaPlayer.cs
mfkl: Putting the dllimport statements in a private struct is a cppsharp convention which I find OK. What advantage would bring using a static class for those?

LibVLCSharp/Shared/MediaPlayer.cs
mfkl: Just to be clear, this is outside the scope of this PR FYI as this API was already like this and this PR is not currently changing it.

> Why do you have a property with only a getter, and a Set...() method? 

Because the setter takes an argument, so it cannot be a traditional C# `get; set;`.

> I'd prefer to have Get...() methods too for symmetry

I.. don't :D and that'd make me Java-ify all the API.

> I'm not really a fan of getters that allocates memory

Well... it releases that memory as well to be fair, and looking into `MediaPlayer` properties at runtime with the debugger seems to be fine.

LibVLCSharp/Shared/MediaPlayer.cs
jeremyVignelles: The only two reasons are :

- `static class` conveys the meaning that it does not name a variable type, but only contains helpers
- I've always done it that way.

Otherwise, if you say that it's a cppsharp convention, I'm fine with this. I don't know the technical reasons over a choice rather than another.

LibVLCSharp/Shared/MediaPlayer.cs
jeremyVignelles: OK, no strong opinions here.

LibVLCSharp/Shared/MediaPlayer.cs
kakone: I also agree with @jeremyVignelles. The small inconvenient with `struct` is that you can instantiate it (you can do `new Native()`, which does not really make sense here).

mfkl: > you can do new Native()

Uh... sure :D but 3rd party users cannot do that, only people editing libvlcsharp.

It's not a very impactful change IMO, but I will think about it and make another PR if we go the `static class` route.

LibVLCSharp/Shared/Helpers/MarshalUtils.cs
mfkl: Done

PR #28, opened 2019-01-22
Add WinForms control and sample

mfkl:
This PR adds a WinForms control so users don't have to do this manually.
This will add an additional nuget package `LibVLCSharp.WinForms` which references `LibVLCSharp`.

To do:

- [x] Figure out why there is no video currently...
- [x] Add csproj info for deployment


mfkl:
Should the `BackColor` set in the Control ctor be customizable?

jeremyVignelles:
Don't think so : VLC draws black anyway as far as I remember

mfkl:
Alright. Will merge soon then.

Samples/LibVLCSharp.WinForms.Sample/Form1.Designer.cs
```diff
@@ -0,0 +1,68 @@
...
+            if (disposing && (components != null))
+            {
+                components.Dispose();
+            }
+            base.Dispose(disposing);
+        }
+
+        #region Windows Form Designer generated code
+
+        LibVLC _libVLC;
```
jeremyVignelles: _libVLC and _mp will need to be declared in ` Form1.cs` because this file needs to be untouched by hand.

Samples/LibVLCSharp.WinForms.Sample/Form1.Designer.cs
```diff
@@ -0,0 +1,68 @@
...
+        MediaPlayer _mp;
+
+        /// <summary>
+        /// Required method for Designer support - do not modify
+        /// the contents of this method with the code editor.
+        /// </summary>
+        private void InitializeComponent()
+        {
+            if(!DesignMode)
+                Core.Initialize();
```
jeremyVignelles: Must be done in the ctor, before calling InitializeComponent

Samples/LibVLCSharp.WinForms.Sample/Form1.Designer.cs
```diff
@@ -0,0 +1,68 @@
...
+
+            _videoView = new VideoView();
+            ((System.ComponentModel.ISupportInitialize)(_videoView)).BeginInit();
+            SuspendLayout();
+
+            _libVLC = new LibVLC("-vv");
+            _mp = new MediaPlayer(_libVLC);
```
jeremyVignelles: these 2 initializations belongs to the ctor

Samples/LibVLCSharp.WinForms.Sample/Form1.Designer.cs
```diff
@@ -0,0 +1,68 @@
...
+            Load += Form1_Load;
+
+            AutoScaleDimensions = new System.Drawing.SizeF(6F, 13F);
+            AutoScaleMode = System.Windows.Forms.AutoScaleMode.Font;
+            ClientSize = new System.Drawing.Size(800, 450);
+            Text = "LibVLCSharp.WinForms";
+            _videoView.Dock = DockStyle.Fill;
+            Controls.Add(_videoView);
+            _videoView.MediaPlayer = _mp;
```
jeremyVignelles: Belongs to the ctor, after InitializeComponents() IMO

Samples/LibVLCSharp.WinForms.Sample/Form1.Designer.cs
```diff
@@ -0,0 +1,68 @@
...
+
+            ((System.ComponentModel.ISupportInitialize)(_videoView)).EndInit();
+            ResumeLayout(false);
+        }
+
+        private void Form1_Load(object sender, System.EventArgs e)
```
jeremyVignelles: Belongs to Form1.cs?

Samples/LibVLCSharp.WinForms.Sample/LibVLCSharp.WinForms.Sample.csproj
```diff
@@ -0,0 +1,100 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<Project ToolsVersion="15.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
```
jeremyVignelles: SDK-style project?

mfkl:
Took your remarks into account. Now the embedding fails (new window created), not sure why yet.

jeremyVignelles:
I meant something more like this : 
https://github.com/jeremyVignelles/libvlcsharp/commit/2259c95bd60122833b2fdd5e45270978f7f7efa2

It does work on my machine, though I had to cheat a little: 
The correct procedure to add a VideoView in a WinForms project is:

- Build the LVS.WinForms project
- Reference it in the sample project
- Go to the Form1.cs, right click/open in designer
- open the toolbox, and drag/drop the VideoView to the Form.

This procedure works in a WinForms project (old-style), but not in the MSBuild.Sdk.Extras project (fails with "Défaillance irrémédiable (Exception de HRESULT : 0x8000FFFF (E_UNEXPECTED))"

Maybe we should ask MSBuild.Sdk.Extras why.

mfkl:
rebased on master.

So the last thing to be discussed is that, with the new sdk style csproj _in the sample project_, there is an exception thrown by VS when attempting to drag and drop the VideoView from the toolbox onto the designer.

Since 
- that's about user code, basically the 3rd party sample app that decides to use the new sdk csproj makes the tooling break, 
- that a workaround can be provided (i.e. use older style csproj if you want to use the designer) and
- that will probably get fixed properly with .NET Core 3 supporting WinForms and the tooling updated for that (i.e. 2019), 
- that can always get fixed later and it's not critical enough to delay the release.

I'd say let's ship it as it is :shipit:

PR #29, opened 2019-02-04
Initial Core.Initialize() refactoring

mfkl:
- [x] Document to callers that it may throw
- [x] provide full path of missing native library in exception.
- [x] Load libvlc libraries from specified folder.  This means we now support loading libvlc elsewhere than our libvlc nuget package.
- [x] log stuff.
- [x] Compute and try more possible locations. especially netcoreapp TFM, from `.nuget` location, etc.
- [x] load from custom path on linux
- [x] Test on blank app new projects on **all** targets with current libvlc nuget packages:
  - [x] Android
  - [x] GTK Windows, Linux
  - [x] net40 Windows
  - [x] .NET core sample app Windows, Mac, Linux
  - [x] ASP.NET Core Windows, Mac, Linux
  - [x] Mac cocoa
- [ ] Unit tests
- [x] Add ASP.NET Core compatibility on readme

### DIFF
--- a/LibVLCSharp/Shared/Core.cs
+++ b/LibVLCSharp/Shared/Core.cs
@@ -1,6 +1,9 @@
 ﻿using System;
+using System.Collections.Generic;
 using System.Diagnostics;
 using System.IO;
+using System.Linq;
+using System.Reflection;
 using System.Runtime.InteropServices;
 #if ANDROID
 using Java.Interop;
@@ -36,21 +39,25 @@ namespace LibVLCSharp.Shared
 #endif
         }
 
+#if NET || NETSTANDARD
         static IntPtr _libvlccoreHandle;
         static IntPtr _libvlcHandle;
-
+#endif
         /// <summary>
-        /// Load the native libvlc library (if necessary depending on platform)
+        /// Load the native libvlc library (if necessary, depending on platform)
+        /// <para/> Ensure that you installed the VideoLAN.LibVLC.[YourPlatform] package in your target project
+        /// <para/> This will throw a <see cref="VLCException"/> if the native libvlc libraries cannot be found or loaded.
         /// </summary>
-        /// <param name="appExecutionDirectory">The path to the app execution directory. 
+        /// <param name="libvlcDirectoryPath">The path to the directory that contains libvlc and libvlccore
         /// No need to specify unless running netstandard 1.1, or using custom location for libvlc
+        /// <para/> This parameter is NOT supported on Linux, use LD_LIBRARY_PATH instead.
         /// </param>
-        public static void Initialize(string appExecutionDirectory = null)
+        public static void Initialize(string libvlcDirectoryPath = null)
         {
 #if ANDROID
             InitializeAndroid();
 #elif NET || NETSTANDARD
-            InitializeDesktop(appExecutionDirectory);
+            InitializeDesktop(libvlcDirectoryPath);
 #endif
         }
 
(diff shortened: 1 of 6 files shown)


LibVLCSharp/Shared/Core.cs
```diff
@@ -39,8 +40,10 @@ struct Native
...
         /// <summary>
         /// Load the native libvlc library (if necessary depending on platform)
+        /// <para/> Ensure that you installed the VideoLAN.LibVLC.[YourPlatform] package in your target project
+        /// <para/> This will throw a <see cref="VLCException"/> if the native libvlc libraries cannot be found or loaded.
         /// </summary>
-        /// <param name="appExecutionDirectory">The path to the app execution directory. 
+        /// <param name="appExecutionDirectory">The path to the directory that contains libvlc and libvlccore
         /// No need to specify unless running netstandard 1.1, or using custom location for libvlc
         /// </param>
         public static void Initialize(string appExecutionDirectory = null)
```
jeremyVignelles: Should we rename the variable as well to match the comment? `libraryDirectory`?

LibVLCSharp/Shared/Core.cs
```diff
@@ -64,73 +67,114 @@ static void InitializeAndroid()
...
+#endif
+            // TODO: add more path
+            return paths;
+        }
+
+        static bool Load(string appExecutionDirectory, bool shouldThrow = true, bool log = true)
+        {
+            if (string.IsNullOrEmpty(appExecutionDirectory)) return false;
+
+            Debug.Assert(PlatformHelper.IsWindows || PlatformHelper.IsMac); // for now
```
jeremyVignelles: What does this do on linux? stop execution? If so, I'd prefer a `if(PlatfotmHelper.IsLinux) return;` or something less critical for the app.

jeremyVignelles: I'd prefer to be able to load the library on linux of course, I have a library I built and can certainly help with that code.

LibVLCSharp/Shared/Core.cs
```diff
@@ -64,73 +67,114 @@ static void InitializeAndroid()
...
         }
 
-        //TODO: Add dlopen for UWP, Linux
-        static IntPtr PreloadNativeLibrary(string nativeLibrariesPath, string libraryName)
+        static bool LoadNativeLibrary(string nativeLibraryPath, out IntPtr handle)
         {
-            Debug.WriteLine($"Loading {libraryName}");
-            var libraryPath = Path.Combine(nativeLibrariesPath, libraryName);
+            handle = IntPtr.Zero;
+            Debug.WriteLine($"Loading {nativeLibraryPath}");
```
jeremyVignelles: What's the point of the  `Log` method if you don't use it?

jeremyVignelles:
> Load libvlc libraries from specified folder. This means we now support loading libvlc elsewhere than our libvlc nuget package.

Great news!

>log stuff.

Is Debug.WriteLine() enough? Do we need to / can we integrate this with a `Microsoft.Extensions.Logging.ILogger`. Maybe with another package that would add this feature?

I mean, in the core, we can emit an event/ call a delegate. From another package, we can add a method to the delegate, and add all received message to the ILogger. The user would probably not call `Core.Initialize()` directly, but rather a `IServiceCollection.AddLibVLC()` method to integrate this in his service pipeline.

> Compute and try more possible locations. especially netcoreapp TFM, from `.nuget` location, etc.

Not sure if `.nuget` would be a good fit for auto discovery. Remember how many times you tried to publish a new package that conflicted with a package in `.nuget`. This place is hidden and would likely cause unexpected behaviors. Moreover, you don't control which versions are inside. What if for another project, I need libvlc 4? would this version be picked automatically for that project?

> Add ASP.NET Core compatibility on readme

What would need to be written?

Speaking about ASP.net core compat, I mentioned earlier ILogger and IServiceCollection which are always referenced in ASP.net core projects. Would a `LibVLCSharp.AspNetCore` help, and use `IHostingEnvironment` for folders discovery? Is `AspNetCore` a good name while `ILogger`and `IServiceProvider` can be used in all .net standard projects (provided that the package is referenced)?

mfkl:
> Is Debug.WriteLine() enough? Do we need to / can we integrate this with a `Microsoft.Extensions.Logging.ILogger`. Maybe with another package that would add this feature?

> I mean, in the core, we can emit an event/ call a delegate. From another package, we can add a method to the delegate, and add all received message to the ILogger. The user would probably not call `Core.Initialize()` directly, but rather a `IServiceCollection.AddLibVLC()` method to integrate this in his service pipeline.

For now I think it's enough. The logging strategy has to be defined and implemented but its out of scope from this PR (tracked here https://code.videolan.org/videolan/LibVLCSharp/issues/37). We could use https://www.nuget.org/packages/Microsoft.Extensions.Logging.Abstractions/ indeed since its netstandard2.0, but lets talk about it in the gitlab issue. 

> Not sure if `.nuget` would be a good fit for auto discovery. Remember how many times you tried to publish a new package that conflicted with a package in `.nuget`. This place is hidden and would likely cause unexpected behaviors. Moreover, you don't control which versions are inside. What if for another project, I need libvlc 4? would this version be picked automatically for that project?

Ok so...

`typeof(LibVLC).Assembly.Location` (currently how it's done)
`Assembly.GetExecutingAssembly().Location` (additional path suggested by @srollinet)
where else?

> What would need to be written?

Just add `(including ASP.NET Core)` next to `.NET Core` in https://github.com/videolan/libvlcsharp#supported-platforms

> Speaking about ASP.net core compat, I mentioned earlier ILogger and IServiceCollection which are always referenced in ASP.net core projects. Would a `LibVLCSharp.AspNetCore` help, and use `IHostingEnvironment` for folders discovery?

Even though I think we should support libvlc loading from aspnetcore projects, efforts to support asp.net as a first class citizen should stay very low prio right now IMO. That could change when we have a collection of compelling use cases, and at least some user-demand. But for now I think there are more impactful things we should focus on.

LibVLCSharp/Shared/Core.cs
```diff
@@ -64,73 +67,114 @@ static void InitializeAndroid()
...
             {
-                _libvlcHandle = PreloadNativeLibrary(appExecutionDirectory, $"{Constants.LibraryName}.dylib");
-                if (_libvlcHandle == IntPtr.Zero)
+                var libvlccorePath = Path.Combine(appExecutionDirectory, $"{Constants.CoreLibraryName}{LibraryExtension}");
+                loadResult = LoadNativeLibrary(libvlccorePath, out _libvlccoreHandle);
+                if (!loadResult) 
                 {
-                    throw new VLCException($"Failed to load required native library {Constants.LibraryName}.dylib");
+                    if (log)
+                        Log($"Failed to load required native library {libvlccorePath}");
```
srollinet: Don't know if it is a good thing to log each failed attempt. It makes the logs harder to read because of the additional (false) failures.
Maybe it is clearer to have a single failure log with all the search locations.

srollinet:
> `typeof(LibVLC).Assembly.Location` (currently how it's done)
`Assembly.GetExecutingAssembly().Location` (additional path suggested by @srollinet)
where else?

Maybe the default Windows installation path? (if windows, and with a lower priority)

LibVLCSharp/Shared/Core.cs
```diff
@@ -64,73 +67,114 @@ static void InitializeAndroid()
...
 
-                _libvlcHandle = PreloadNativeLibrary(librariesFolder, $"{Constants.LibraryName}.dll");
-                
-                if(_libvlcHandle == IntPtr.Zero)
-                {
-                    throw new VLCException($"Failed to load required native library {Constants.LibraryName}.dll");
-                }
+            if(!Loaded)
+            {
+                throw new VLCException($"Failed to load required native libraries");
```
srollinet: It is maybe useful to put all the search locations in this exception

jeremyVignelles:
@srollinet : it's already planned to look into the default locations (program files on windows, /usr/bin (not sure) on linux...)

jeremyVignelles:
> and at least some user-demand.

Now, we have one 🙂 :

https://code.videolan.org/videolan/LibVLCSharp/issues/37#note_29242

mfkl:
>  it's already planned to look into the default locations (program files on windows, /usr/bin (not sure) on linux...)

No it's not :) For Windows I am not a fan _at all_ of this practice and don't intend on looking at that location. 
- It makes deployment a pain (users need to install VLC? And the correct version from the official website...).
- What if the dev updates VLC on his pc? That'd update libvlc for all this apps.
That's why we have https://github.com/mfkl/libvlc-nuget in the first place. For Linux, for now [it's not clear](https://code.videolan.org/videolan/LibVLCSharp/issues/97) what the best course of action is, but mono looks in `usr/bin` anyway so no need for additional paths for now.

jeremyVignelles:
Oh, I thought so.  You're probably right.

LibVLCSharp/Shared/Core.cs
```diff
@@ -64,75 +69,135 @@ static void InitializeAndroid()
...
-            else if (PlatformHelper.IsLinux)
+
+            if (!Loaded)
             {
-                // Initializes X threads before calling VLC. This is required for vlc plugins like the VDPAU hardware acceleration plugin.
-                if (Native.XInitThreads() == 0)
-                {
-#if !NETSTANDARD1_1
-                    Trace.WriteLine("XInitThreads failed");
+                throw new VLCException($"Failed to load required native libraries. Search paths include {paths.ToString()}");
```
srollinet: `paths.ToString()` won't return what you want here. It will return the type name of `paths`.
You have to do something like that `string.Join(",", paths.Select(p => $"{p.libvlc},{p.libvlccore}"));`

mfkl: Yes you're right. I suspected it, but wanted to wait until I test to get the formatting right. Thanks for the heads up though.

mfkl:
Merged.

PR #34, opened 2019-03-06
Fix more UTF8 interop calls

mfkl:
- remove Utf8Marshaler. It's not used because netstandard1.1 support would mean #if #def'ing all dll import statements.
- free all native strings allocated on our side after the native calls.
- free all the libvlc strings that needs freeing (`char *` returns).
```
libvlc_video_get_aspect_ratio
libvlc_video_get_crop_geometry
libvlc_video_get_marquee_string
libvlc_audio_output_device_get (next PR)
libvlc_media_get_mrl
libvlc_media_get_meta
```
- Use extension method everywhere (looks cleaner)


LibVLCSharp/Shared/Media.cs
```diff
@@ -299,7 +311,7 @@ public void AddOption(MediaConfiguration mediaConfiguration)
         }
 
         /// <summary>Add an option to the media with configurable flags.</summary>
-        /// <param name="options">the options (as a string)</param>
+        /// <param name="options">the option (as a string)</param>
```
jeremyVignelles: you didn't change name= here, but you changed the parameter name below

mfkl:
> new assymetric way

What do you mean by that? What would you prefer doing?

jeremyVignelles:
I mean that the `ToUtf8()` method allocates a buffer, which is released elsewhere in the code (with no guarantee that the free is actually done).

Vlc.DotNet uses a SafeHandle for that, and you "know" that a IDisposable should be disposed or wrapped into a using.

On the other hand, `FromUtf8` has the ability to free the input by itself.

mfkl:
> which is released elsewhere in the code (with no guarantee that the free is actually done).

It is [documented](https://github.com/videolan/libvlcsharp/pull/34/files#diff-c21ba0f069016164a883a740fe9ba5fbR115) though. You'd rather we used a `SafeHandle` like in Vlc.DotNet?

jeremyVignelles:
I'd rather use the new marshalling capabilities but since we need to support older versions... SafeHandle would be better IMO, but no strong feelings though

PR #36, opened 2019-03-08
Improves MVVM scenario

kakone:
PR to handle this [MVVM scenario](https://github.com/kakone/libvlcsharp_samples/blob/master/Samples/Samples/MvvmSampleViewModel.cs) where all the code is in the ViewModel (the [MainPage.cs](https://github.com/kakone/libvlcsharp_samples/blob/master/Samples/Samples/MvvmSamplePage.xaml.cs) has no code).

As MediaPlayerChanged event is called too early in this case, I needed to add this code in the VideoViewRenderers : 
```
if (Control.MediaPlayer != e.NewElement.MediaPlayer)
{
  OnMediaPlayerChanging(this, new MediaPlayerChangingEventArgs(Control.MediaPlayer, e.NewElement.MediaPlayer));
}
```

This PR also removed the Loaded event in the VideoView, replaced by a MediaPlayerChanged event. I use now the PropertyChanging event to set the MediaPlayer property in the native control. So, when MediaPlayerChanged event is raised, all is well set (so the HWND for WPF is well defined in the MediaPlayerChanged event).

### DIFF
--- a/Samples/Forms/LibVLCSharp.Forms.Sample/MainPage.xaml.cs
+++ b/Samples/Forms/LibVLCSharp.Forms.Sample/MainPage.xaml.cs
@@ -8,11 +8,5 @@ namespace LibVLCSharp.Forms.Sample
         {
             InitializeComponent();
         }
-
-        protected override void OnAppearing()
-        {
-            base.OnAppearing();
-            ((MainViewModel)BindingContext).OnAppearing();
-        }
     }
 }
--- a/Samples/Forms/LibVLCSharp.Forms.Sample/MainPage.xaml.cs
+++ b/Samples/Forms/LibVLCSharp.Forms.Sample/MainPage.xaml.cs
@@ -9,22 +9,22 @@ namespace LibVLCSharp.Forms.Sample
         {
             InitializeComponent();
             videoView.MediaPlayerChanged += VideoView_MediaPlayerChanged;
-            BindingContext.PropertyChanged += BindingContext_PropertyChanged;
-            BindingContext.Play();
+            ViewModel.PropertyChanged += BindingContext_PropertyChanged;
+            ViewModel.Play();
         }
 
-        private new MainViewModel BindingContext => (MainViewModel)base.BindingContext;
+        private MainViewModel ViewModel => (MainViewModel)BindingContext;
 
         private void BindingContext_PropertyChanged(object sender, System.ComponentModel.PropertyChangedEventArgs e)
         {
-            if (e.PropertyName.Equals(nameof(BindingContext.MediaPlayer)))
+            if (e.PropertyName.Equals(nameof(ViewModel.MediaPlayer)))
                 Debug.WriteLine("MediaPlayer change raised from ViewModel.Propertychanged");
         }
 
         private void VideoView_MediaPlayerChanged(object sender, LibVLCSharp.Shared.MediaPlayerChangedEventArgs e)
         {
             Debug.WriteLine("VideoView_MediaPlayerChanged");
-            BindingContext.Play();
+            ViewModel.Play();
         }
     }
 }
--- a/Samples/Forms/LibVLCSharp.Forms.Sample/MainPage.xaml.cs
+++ b/Samples/Forms/LibVLCSharp.Forms.Sample/MainPage.xaml.cs
@@ -1,5 +1,4 @@
-﻿using System.Diagnostics;
-using Xamarin.Forms;
+﻿using Xamarin.Forms;
 
 namespace LibVLCSharp.Forms.Sample
 {
@@ -8,23 +7,12 @@ namespace LibVLCSharp.Forms.Sample
         public MainPage()
         {
             InitializeComponent();
-            videoView.MediaPlayerChanged += VideoView_MediaPlayerChanged;
-            ViewModel.PropertyChanged += BindingContext_PropertyChanged;
-            ViewModel.Play();
         }
 
-        private MainViewModel ViewModel => (MainViewModel)BindingContext;
-
-        private void BindingContext_PropertyChanged(object sender, System.ComponentModel.PropertyChangedEventArgs e)
+        protected override void OnAppearing()
         {
-            if (e.PropertyName.Equals(nameof(ViewModel.MediaPlayer)))
-                Debug.WriteLine("MediaPlayer change raised from ViewModel.Propertychanged");
-        }
-
-        private void VideoView_MediaPlayerChanged(object sender, LibVLCSharp.Shared.MediaPlayerChangedEventArgs e)
-        {
-            Debug.WriteLine("VideoView_MediaPlayerChanged");
-            ViewModel.Play();
+            base.OnAppearing();
+            ((MainViewModel)BindingContext).OnAppearing();
         }
     }
 }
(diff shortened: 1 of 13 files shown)


Samples/Forms/LibVLCSharp.Forms.Sample/MainPage.xaml.cs
```diff
@@ -1,44 +1,30 @@
...
-            _vm.Initialize();
+            BindingContext.PropertyChanged += BindingContext_PropertyChanged;
+            BindingContext.Play();
         }
 
-        private void VideoView_Loaded(object sender, System.EventArgs e)
-        {
-            _vm.MediaPlayer.Play();
-        }
+        private new MainViewModel BindingContext => (MainViewModel)base.BindingContext;
```
jeremyVignelles: Not a huge fan of `new` overrides. could we call that another way? (I tend to call that typed property `ViewModel`)

Samples/Forms/LibVLCSharp.Forms.Sample/MainViewModel.cs
```diff
@@ -1,12 +1,20 @@
...
 namespace LibVLCSharp.Forms.Sample
 {
     public class MainViewModel : INotifyPropertyChanged
     {
         public event PropertyChangedEventHandler PropertyChanged;
-        LibVLC _libVLC;
+
+        public MainViewModel()
+        {
+            Task.Run((Action)Initialize);
```
jeremyVignelles: Seems like a hack and can't handle exceptions properly. What's the issue with calling Initialize externally?

mfkl:
> As MediaPlayerChanged event is called too early in this case

Yes, that's expected and documented behavior https://github.com/videolan/libvlcsharp/blob/8c24dff1348c19e68ace1abecbff37b06b581f2d/LibVLCSharp.Forms/Shared/VideoView.cs#L18

> I needed to add this code in the VideoViewRenderers :
```
if (Control.MediaPlayer != e.NewElement.MediaPlayer)
{
  OnMediaPlayerChanging(this, new MediaPlayerChangingEventArgs(Control.MediaPlayer, e.NewElement.MediaPlayer));
}
```

This code allows us to remove the requirement of listening to `VideoView.Loaded` and that's great. Will test later :)

I used `Loaded` to follow the usual .NET API conventions, don't you think that it makes sense to keep it?
Can you think of a use case where the event `MediaPlayerChanging` is useful to users?

Samples/Forms/LibVLCSharp.Forms.Sample/MainPage.xaml.cs
kakone: OK, I'm going to rename it to ViewModel.

mfkl:
By telling the users to only call Play() after/when `Loaded` has been raised, we avoided having a mediaplayer without hwnd that started playing and creating a VLC d3d11 output window by itself.

With your code we'd not give that guideline to our users anymore. Can this case still happen?

Samples/Forms/LibVLCSharp.Forms.Sample/MainViewModel.cs
kakone: No, it's not a hack, it's to show that we can handle a true MVVM scenario. All the code (initialization, play, ...) can be done in the ViewModel. I kept some code in the MainPage.xaml.cs file but it' only to handle the WPF case (where Play method must be called after that the HWND is set).
There is no problem to call the Initialize method externally but it breaks the MVVM scenario.
For the exceptions, you can handle them in the Initialize method, what's wrong ?

kakone:
> I used Loaded to follow the usual .NET API conventions, don't you think that it makes sense to keep it?

The problem with the Loaded event is that it only worked for the first assignment. If you updated the MediaPlayer property, it didn't work.

> Can you think of a use case where the event MediaPlayerChanging is useful to users?

No, I think it's only useful internally.

> By telling the users to only call Play() after/when Loaded has been raised, we avoided having a mediaplayer without hwnd that started playing and creating a VLC d3d11 output window by itself.
> 
> With your code we'd not give that guideline to our users anymore. Can this case still happen?

It's still happens. If the user uses the MediaPlayer property of the VideoView, there is no problem but in other cases (MVVM, ...), the user should wait that the MediaPlayer property of the VideoView is well set (so, that the HWND is set) before call the Play method.

mfkl:
>  Can you think of a use case where the event MediaPlayerChanging is useful to users?
> No, I think it's only useful internally.

Then it should be set to `internal` (https://github.com/videolan/libvlcsharp/pull/36/files#diff-20a000518866fe42f5225ea0bb0e2018R13)

kakone:
> Then it should be set to internal

I can't set it to internal because Forms.Platforms.GTK and Forms.Platforms.WPF projects use it.

mfkl:
You're missing a `using LibVLCSharp.Shared;` in `LibVLCSharp.Forms.Platforms.GTK/VideoViewRenderer.cs`

kakone:
Yes, fixed.

mfkl:
About 50% of the time, I get the window created outside of the app with the Xamarin.Forms.WPF sample.



Not sure if that timing issue got introduced with your changes or if it was already there.

kakone:
I don't reproduce the problem but I think I found a good fix for this WPF problem. If I create the LibVLC instance with the option `--no-embedded-video`, this window will be never created and we can set the HWND later.

Perhaps, we must activate this option automatically in the LibVLC constructor ?

mfkl:
I try to avoid using libvlc CLI options in LibVLCSharp as much as possible. For instance, I recently worked on an app that actually needed VLC to create those win32 windows. As long as play is not called before setHwnd has returned, it should work, so we should find a way to fix this?

kakone:
> As long as play is not called before setHwnd has returned, it should work, so we should find a way to fix this?

That's exactly the problem for me. We shouldn't have to the call Play method only after setting the Hwnd. The developer should have the choice to set it when he wants, before or after that the media starts to play.

kakone:
By setting `--no-embedded-video` by default in the LibVLC constructor, the developper will still be able to create those win32 windows: he can pass the option `embedded-video` when he creates the LibVLC instance.

kakone:
I updated the Forms.Sample in order to make it work with WPF. I must wait that the VideoView is well shown and the MediaPlayer property of the VideoView is set.

mfkl:
Just tried your code:
- The WPF.Forms sample seems to be working 100% of the time. There are a few glitches at the beginning though. Probably related to the XF.WPF theme.
- The regular WPF sample seems to be broken 100% of the time. `No MediaPlayer is set, aborting...` probably a timing issue.

kakone:
> The regular WPF sample seems to be broken 100% of the time. No MediaPlayer is set, aborting... probably a timing issue.

I didn't modify the native control, nor the WPF sample. The message `No MediaPlayer is set, aborting...` is normal, the MediaPlayer property is set later (and the MediaPlayer.Hwnd property is well set in the OnMediaPlayerChanged method). When you press "Play"button, it works well.

mfkl:
> When you press "Play"button, it works well.

You're saying the WPF sample works for you? I just tried on both master and your branch (rebased on master) and it doesn't work (both branches). The example 1 and example 2 window just load blank, no button to press.

kakone:
Yes, it works. The loaded window have two buttons PLAY and STOP at bottom :

image

mfkl:
I know what it's supposed to look like, those buttons just aren't there on my machine (anymore) for some reason... guess that's ready to merge then.

mfkl:
so the reason I could not see the buttons is because


Which seems to get fixed by https://github.com/videolan/libvlcsharp/pull/42

mfkl:
Merged https://github.com/videolan/libvlcsharp/commit/332cf8d085583440f7fb234cadb0650ff8e6d45b

PR #39, opened 2019-03-23
apply dotnet format and add code analyzers

hanabi1224:
apply [dotnet-format](https://github.com/dotnet/format) and add code analyzers around threading and disposables


mfkl:
Hi @hanabi1224,

Your PR seems to attempt 3 different things, so I think it'd be better if you split it in 3 commits.
- Add .editorconfig and related formatting changes. You should indicate in the commit message that it closes https://code.videolan.org/videolan/LibVLCSharp/issues/121
- Disposable base class: this has been attempted before, and reverted. See https://code.videolan.org/videolan/LibVLCSharp/merge_requests/2 and https://code.videolan.org/videolan/LibVLCSharp/commit/e3cb50a7ca062e950f8545f02387ac646ced7c10 as to why. 
- Code analyzers. Besides the package references, your IDisposable changes relate to the IDisposable analyzer. I don't see any changes related to the Threading/async analyzers. Also `Microsoft.Net.Compilers` is apache2, which is usually a no-go for us (though this is a dev only dependency).

hanabi1224:
@mfkl Thanks for your reply, I have created a separate PR https://github.com/videolan/libvlcsharp/pull/40 to address bullet 1

PR #42, opened 2019-03-31
Fix Foreground location when using High-DPI

wjax:
Fix for HighDPI Screens.

### DIFF
--- a/LibVLCSharp.WPF/ForegroundWindow.xaml.cs
+++ b/LibVLCSharp.WPF/ForegroundWindow.xaml.cs
@@ -38,7 +38,7 @@ namespace LibVLCSharp.WPF
         void Background_Unloaded(object sender, RoutedEventArgs e)
         {
             _wndhost.Closing -= Wndhost_Closing;
-            _wndhost.SizeChanged -= Wndhost_SizeChanged;
+            _bckgnd.SizeChanged -= Wndhost_SizeChanged;
             _wndhost.LocationChanged -= Wndhost_LocationChanged;
 
             Hide();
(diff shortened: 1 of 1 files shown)


jeremyVignelles:
Hi, can you be more verbose about what your issue was? is that related to what you said on Vlc.DotNet's gitter?

wjax:
Hi,

When using current solution on a High DPI screen like the one in Surface Book, the Foreground window is no well sized.
So I modified and tested the proposed solution.

I will try to upload some screenshots later.

wjax:
Here you have.
Before and after my pull request

The cross is supposed to be centered.

image

image

wjax:
This is the UserControl that is set as Content of VideoView
```
<UserControl x:Class="LibVLCSharp.WPF.Sample.OSDMain"
             xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
             xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
             xmlns:local="clr-namespace:LibVLCSharp.WPF.Sample"
             mc:Ignorable="d" 
             d:DesignHeight="450" d:DesignWidth="800">
    <Viewbox>
        <Canvas Width="1000" Height="1000">
            <Line X1="450" X2="550" Y1="500" Y2="500" Stroke="Black" StrokeThickness="2"/>
            <Line X1="500" X2="500" Y1="450" Y2="550" Stroke="Black" StrokeThickness="2"/>
        </Canvas>
    </Viewbox>
</UserControl>
```

mfkl:
Thanks! Merged https://github.com/videolan/libvlcsharp/commit/2dc77dbb2de51f054d499aab64ea95c45f4bdd0e

PR #43, opened 2019-04-01
UWP support

mfkl:
See https://code.videolan.org/videolan/LibVLCSharp/issues/116

PR #44, opened 2019-04-06
Implemented UWP platform

jeremyVignelles:
This replaces #43 .

This is a new implementation of LibVLCSharp for UWP, with a VideoView, like the other targets.

Also added a Sample project to demonstrate how it works.


LibVLCSharp/LibVLCSharp.csproj
```diff
@@ -62,9 +63,14 @@
...
   </ItemGroup>
   <ItemGroup Condition=" '$(TargetFramework)' == 'net40' ">
-    <PackageReference Include="System.ValueTuple" Version="4.5.0"/>
+    <PackageReference Include="System.ValueTuple" Version="4.5.0" />
     <PackageReference Include="Microsoft.Bcl.Async" Version="1.0.168" />
   </ItemGroup>
+  <ItemGroup Condition="$(TargetFramework.StartsWith('uap'))">
+    <Compile Include="Platforms\UWP\**\*.cs" />
+    <PackageReference Include="System.ValueTuple" Version="4.5.0" />
+    <PackageReference Include="SharpDX.Direct3D11" Version="4.2.0" />
```
jeremyVignelles: For further reference, SharpDX has just been announced as not maintained anymore. We chose to use it anyway, given the maturity of the project. Maybe that decision will be reconsidered later.

mfkl: Yes.

Samples/LibVLCSharp.UWP.Sample/MainPage.xaml.cs
```diff
@@ -0,0 +1,34 @@
...
+        private LibVLC _libVLC;
+
+        private MediaPlayer _mediaPlayer;
+
+        public MainPage()
+        {
+            InitializeComponent();
+            Loaded += (s, e) =>
+            {
+                _libVLC = new LibVLC(VideoView.GetSwapChainOptions());
```
mfkl: `GetSwapChainOptions` is an implementation detail; it cannot be customized and the users do not need to know about it. I think it should be `internal` and injected automatically in the LibVLC ctor.

LibVLCSharp/Platforms/UWP/VideoView.cs
```diff
@@ -0,0 +1,233 @@
...
+                SwapEffect = SwapEffect.FlipSequential,
+                Flags = SwapChainFlags.None,
+                AlphaMode = AlphaMode.Unspecified
+            };
+
+            _swapChain = new SharpDX.DXGI.SwapChain1(dxgiFactory, _d3d11Device, ref swapChainDescription);
+
+            _device.MaximumFrameLatency = 1;
+
+            //TODO: perform the next 2 calls on the UI thread
```
jeremyVignelles: This is a legacy of the original code. What we can do:

- Leave as-is, and we need to investigate later
- We are already on the main thread because we handled the Loaded event, and this is fine, and we can remove the comment
- All the other calls must be done on another thread, and maybe we should switch threads here

LibVLCSharp/Platforms/UWP/VideoView.cs
```diff
@@ -0,0 +1,233 @@
...
+        SwapChain1 _swapChain;
+
+        bool _loaded;
+
+        /// <summary>
+        /// The constructor
+        /// </summary>
+        public VideoView()
+        {
+            this._panel = new SwapChainPanel();
```
mfkl: Please remove the `this.` (yes I know we need that editorconfig badly).

LibVLCSharp/Platforms/UWP/VideoView.cs
```diff
@@ -0,0 +1,233 @@
...
+            Marshal.FreeHGlobal(width);
+            Marshal.FreeHGlobal(height);
+        }
+
+        void UpdateScale()
+        {
+            _swapchain2.MatrixTransform = new RawMatrix3x2 { M11 = 1.0f / this._panel.CompositionScaleX, M22 = 1.0f / this._panel.CompositionScaleY };
+        }
+
+        void Trim()
```
jeremyVignelles: Oh, I forgot to call that. We need to do that when the app is suspending.

LibVLCSharp/Platforms/UWP/VideoView.cs
```diff
@@ -0,0 +1,233 @@
...
+            _device.MaximumFrameLatency = 1;
+
+            //TODO: perform the next 2 calls on the UI thread
+            var panelNative = ComObject.As<ISwapChainPanelNative>(this._panel);
+            panelNative.SwapChain = _swapChain;
+
+            // This is necessary so we can call Trim() on suspend
+            this._device3 = _device.QueryInterface<SharpDX.DXGI.Device3>();
+            if (_device3 == null)
+                throw new Exception();
```
mfkl: I added this while debugging during the implementation. Probably we can remove it (or add a message if we keep it)

LibVLCSharp/Platforms/UWP/VideoView.cs
```diff
@@ -0,0 +1,233 @@
...
+            this._swapchain2 = _swapChain.QueryInterface<SharpDX.DXGI.SwapChain2>();
+            if (_swapchain2 == null)
+                throw new Exception();
```
mfkl: I added this while debugging during the implementation. Probably we can remove it (or add a message if we keep it)

LibVLCSharp/Platforms/UWP/VideoView.cs
```diff
@@ -0,0 +1,233 @@
...
+
+            UpdateScale();
+            UpdateSize();
+            this._loaded = true;
+        }
+
+        void OnUnload()
```
mfkl: we need to test consecutive unloads with the same LibVLC.cs

LibVLCSharp/Platforms/UWP/VideoView.cs
mfkl: What about this? Have we decided how to get this called on OnSuspending?

LibVLCSharp/Platforms/UWP/VideoView.cs
```diff
@@ -0,0 +1,233 @@
...
+    using SharpDX.Mathematics.Interop;
+    using System;
+    using System.Runtime.InteropServices;
+    using Windows.UI.Xaml.Controls;
+    using Device3 = SharpDX.DXGI.Device3;
+
+    /// <summary>
+    /// The VideoView implementation for UWP applications
+    /// </summary>
+    public class VideoView : UserControl, IVideoView
```
mfkl: Should this inherit from `SwapChainPanel`?

LibVLCSharp/Platforms/UWP/VideoView.cs
mfkl: > We are already on the main thread because we handled the Loaded event, and this is fine, and we can remove the comment

Our users might not run this code from the Loaded event. Unsure if that's an issue here...

LibVLCSharp/Platforms/UWP/VideoView.cs
mfkl: Yeah. Don't know if we can do that from library code without custom user callback

LibVLCSharp/Platforms/UWP/VideoView.cs
jeremyVignelles: Nope this is our code called on load that creates the swapchain, and before the user actually creates the libvlc instance

LibVLCSharp/Platforms/UWP/VideoView.cs
jeremyVignelles: The panel is inside here, not sure if we should change that and make it a panel itself. a user control would be more flexible for the future

Samples/LibVLCSharp.UWP.Sample/MainPage.xaml.cs
jeremyVignelles: It can't because the LibVLC instance doesn't know about the video view nor the swapchain...

mfkl: Hence `Document that a swapchain is required in the LibVLC ctor for UWP builds`

LibVLCSharp/Platforms/UWP/VideoView.cs
jeremyVignelles: We can. Done.

LibVLCSharp/Platforms/UWP/VideoView.cs
jeremyVignelles: now throws a VLCException

LibVLCSharp/Platforms/UWP/VideoView.cs
jeremyVignelles: The method is now called `DestroySwapChain`, but is still private : The VideoView manages completely when this is called. Two consecutive unloads isn't a problem since we always check for nulls before Disposing.

Should we call `DestroySwapChain` at the beginning of the `CreateSwapChain` method? I don't think so for the same reason : it's a private member called on the `Loaded` event

LibVLCSharp/Platforms/UWP/VideoView.cs
jeremyVignelles: This is now called.

jeremyVignelles:
On the UWP.Sample, I'm getting this warning when building:

>NU1603: LibVLCSharp.UWP.Sample depends on Microsoft.NETCore.UniversalWindowsPlatform (>= 6.2.7) but Microsoft.NETCore.UniversalWindowsPlatform 6.2.7 was not found. An approximate best match of Microsoft.NETCore.UniversalWindowsPlatform 6.2.8 was resolved.

Anyone has a clue of what that means?

Samples/LibVLCSharp.UWP.Sample/MainPage.xaml.cs
jeremyVignelles: Documented

jeremyVignelles:
> I was thinking ` [API diff] Document that a swapchain is required in the LibVLC ctor for UWP builds` from https://code.videolan.org/videolan/LibVLCSharp/issues/116 but looks like its different here. Thoughts?

This is exactly what I've done : pass a swapchain in the LibVLC constructor. This is documented

>     * We need to add a UWP custom renderer for Forms support.

Forms support will be made in a separate PR as discussed together since it is not as trivial as it seems. I'm in favor of implementing Forms.UWP support in V4 only.

>     * Docs, csproj description and readme will need an update as well.

Updated.

LibVLCSharp/Platforms/UWP/VideoView.cs
jeremyVignelles: From the beginning of the method to this line, the code takes 23ms on my computer to run.

Would switching threads bring any benefit there? I think that would only make the code harder to read. I'll remove the //TODO.

README.md
```diff
@@ -61,6 +61,7 @@ Mono, .NET Framework and .NET Core runtimes are supported.
 - Xamarin.tvOS
 - Xamarin.Mac (Cocoa)
 - Windows (WPF, WinForms, GTK)
+- Windows 10 (UWP)
```
mfkl: I think it'd be great to mention mobile, desktop and Xbox.

mfkl:
https://github.com/videolan/libvlcsharp/blob/0bb271e9097af096b483803895af42e3294213c4/LibVLCSharp/Shared/Core.cs#L43 Afraid we need yet another `#if` to exclude `_libvlccoreHandle` from uap builds to avoid the build warning.

LibVLCSharp/Platforms/UWP/VideoView.cs
```diff
@@ -0,0 +1,286 @@
...
+        }
+
+        /// <summary>
+        /// Gets the swapchain parameters to pass to the <see cref="LibVLC"/> constructor.
+        ///
+        /// If you don't pass them to the <see cref="LibVLC"/> constructor, the video won't
+        /// be displayed in your application.
+        /// </summary>
+        /// <returns>The list of arguments to be given to the <see cref="LibVLC"/> constructor.</returns>
+        public string[] GetSwapChainOptions()
```
mfkl: Why not make this a property `SwapChainOptions` with a getter?

mfkl:
Merged https://github.com/videolan/libvlcsharp/commit/1bf388e316322f0be0618f578011cfb44bbb33d8. Great job @jeremyVignelles

PR #46, opened 2019-04-11
Code formatter on appveyor

hanabi1224:
Here's a PR to enforce code format with appveyor CI,  it levarages 'dotnet format --check --dry-run -w scripts' which returns non 0 exit code when there're files to be formatted 

Here's a build to demonstrate build failure: https://ci.appveyor.com/project/hanabi1224/libvlcsharp/builds/23775799

And here's a successful build after applying format changes: https://ci.appveyor.com/project/hanabi1224/libvlcsharp/builds/23776112

@jeremyVignelles  @mfkl


mfkl:
Hi @hanabi1224,

Sorry but I can't take those changes.

- There must be a way to format the code without adding an additional csproj, which would be hard to maintain and confuse users.
- We're using Azure DevOps for CI, I don't think it's necessary to use another service just for formatting.

PR #48, opened 2019-05-15
Fix null ref on iOS with X.F 4.0

mfkl:
Fixes https://code.videolan.org/mfkl/libvlcsharp-samples/issues/13

Need a sample with navigation to test "back", like the gestures one https://code.videolan.org/mfkl/libvlcsharp-samples/tree/master/Gestures/Gestures

Tested on:
- [x] iOS
- [x] Android
- [ ] macOS


mfkl:
Looks like the macOS backend has been broken for a bit.

mfkl:
Merged. Will fix macOS XF separately as it's an unrelated issue

PR #52, opened 2019-06-13
Add MediaElement initial setup

mfkl:
This work has been done by @kakone [here](https://github.com/kakone/LibVLCSharp.Forms.MediaPlayerElement/) and intend to fill the need for a customizable, data-binded, Xamarin.Forms MediaElement control https://code.videolan.org/videolan/LibVLCSharp/issues/45

The goal here is to get the code out there _now_, so we can gather feedback. It is not to make the perfect API or make long lasting decisions. Review feedback is welcome still.

- This will be initially published as a **pre-release** LibVLCSharp.Forms package, so API is not stable. Expect API changes.
- We will fix the bugs and issues that arise, but they are expected.
- More work is needed regarding the features and platform support (currently only iOS/Android). Help welcome.


mfkl:
Any reason you used the `Material Icons font` instead of the regular system icons @kakone? Think that the licensing might be an issue for us.

kakone:
I used the `Material Icons font` because there were all the icons I needed, it was the most convenient for me. For example, cast or closed captions icons are missing in the default drawables. 
The icon font should not be distributed in the NuGet package, so I didn't think it would be an issue.

jeremyVignelles:
@mfkl : I see a lot of commits that are totally unrelated. Did you intend to make the PR against 3.x instead of master?

LibVLCSharp.Forms/Shared/PlaybackControls.xaml.cs
```diff
@@ -0,0 +1,1384 @@
...
+
+                UpdateSeekBar(timeSpanPosition, pos);
+            }
+            catch (Exception ex)
+            {
+                ShowErrorMessageBox(ex);
+            }
+        }
+
+        private void UpdateTime(double? position = null)
```
jeremyVignelles: I don't understand this method, and how does it compare with `UpdatePosition`? How come doesn't it take a time as an argument? In my experience, time is quite always set, while position might be unknown (because length can be unknown). This can happen on live streams, and maybe on big files that are long to download?

LibVLCSharp.Forms/Shared/Resources/Strings.fr.resx
```diff
@@ -0,0 +1,141 @@
...
+    <value>Caster sur</value>
+  </data>
+  <data name="ClosedCaptions" xml:space="preserve">
+    <value>Sous-titres</value>
+  </data>
+  <data name="Disable" xml:space="preserve">
+    <value>Désactiver</value>
+  </data>
+  <data name="Error" xml:space="preserve">
+    <value>Error</value>
```
jeremyVignelles:
```suggestion
    <value>Erreur</value>
```

I assume it's an error 😄

LibVLCSharp.Forms/Shared/Themes/Generic.xaml
```diff
@@ -0,0 +1,212 @@
...
+                    x:Class="LibVLCSharp.Forms.Shared.Themes.Generic">
+    <converters:BufferingProgressToBoolConverter x:Key="BufferingProgressToBoolConverter" />
+    <converters:ObjectToBoolConverter x:Key="ObjectToBoolConverter" />
+
+    <Color x:Key="ForeColor">White</Color>
+    <Color x:Key="MainColor">#E57A00</Color>
+    <Color x:Key="ButtonColor">#A0E57A00</Color>
+
+    <OnPlatform x:Key="IconFontFamily" x:TypeArguments="x:String"
+                Android = "MaterialIcons-Regular.ttf#Material Icons"
```
jeremyVignelles: I didn't understand the discussion here. Does it need to be redistributed somehow by the app?

LibVLCSharp.Forms/Shared/PlaybackControls.xaml.cs
```diff
@@ -0,0 +1,1384 @@
...
+            {
+                await controlsPanel.FadeTo(1);
+            }
+            if (FadeOutEnabled && ShowAndHideAutomatically && MediaPlayer?.State == VLCState.Playing)
+            {
+                FadeOutTimer.Change(TimeSpan.FromSeconds(3), TimeSpan.FromMilliseconds(-1));
+            }
+        }
+    }
+}
```
jeremyVignelles: Can we reduce the file size/move things outside of this file?

Sample/Sample.Android/Assets/AboutAssets.txt
```diff
@@ -0,0 +1,19 @@
+Any raw assets you want to be deployed with your application can be placed in
```
jeremyVignelles: What is this `Sample` folder?

Samples/Forms/LibVLCSharp.Forms.MediaElement/LibVLCSharp.Forms.MediaElement.Sample/MainViewModel.cs
```diff
@@ -0,0 +1,68 @@
...
+        {
+            Core.Initialize();
+
+            LibVLC = new LibVLC();
+
+            var media = new Media(LibVLC,
+                "http://download.blender.org/peach/bigbuckbunny_movies/big_buck_bunny_480p_surround-fix.avi",
+                FromType.FromLocation);
+
+            if(Device.RuntimePlatform == Device.Android)
```
jeremyVignelles: Wasn't the purpose of `MediaConfiguration.EnableHardwareDecoding` to be cross-platform and not requiring such tests? Can we remove it here?

Samples/Forms/LibVLCSharp.Forms.MediaElement/LibVLCSharp.Forms.MediaElement.Sample/Sample.xml
```diff
@@ -0,0 +1,58 @@
+<?xml version="1.0"?>
```
jeremyVignelles: Looks like a generated file in the wrong place

mfkl:
> How much effort would be needed to make a desktop Xamarin.Forms implementation in the sample? GTK#? UWP? WPF seems hard to do...

GTK# probably possible.
UWP we'd need to do bring XF support first.
WPF I tried to make it work, but there are issues (and some dependencies that are not compatible).

LibVLCSharp.Forms/Shared/Themes/Generic.xaml
mfkl: Yes

Samples/Forms/LibVLCSharp.Forms.MediaElement/LibVLCSharp.Forms.MediaElement.Sample/MainViewModel.cs
mfkl: Yes

Samples/Forms/LibVLCSharp.Forms.MediaElement/LibVLCSharp.Forms.MediaElement.Sample/Sample.xml
kakone: You are right, it must be deleted because it is generated into the bin folder.

LibVLCSharp.Forms/Shared/Converters/BufferingProgressToBoolConverter.cs
```diff
@@ -0,0 +1,33 @@
...
+    public class BufferingProgressToBoolConverter : IValueConverter
+    {
+        /// <summary>
+        /// Modifies the source data before passing it to the target for display in the UI.
+        /// </summary>
+        /// <param name="value">The source data being passed to the target.</param>
+        /// <param name="targetType">The type of the target property.</param>
+        /// <param name="parameter">An optional parameter to be used in the converter logic.</param>
+        /// <param name="culture">The culture of the conversion.</param>
+        /// <returns>true if value is not equals to 0, false otherwise</returns>
```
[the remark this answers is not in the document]
kakone: OK, it returns true if the value is not equals to 0 and 1, false otherwise. The goal is to hide the progress bar when the buffering progress value is equal to 0 or 1 (so to show it when the progress value is beetween 0 and 1).

LibVLCSharp.Forms/Shared/MediaPlayerElement.xaml.cs
```diff
@@ -0,0 +1,210 @@
...
+        }
+
+        private void PageAppearing(object sender, Page e)
+        {
+            if (e == this.FindAncestor<Page>())
+            {
+                MessagingCenter.Subscribe<LifecycleMessage>(this, "OnSleep", m =>
+                {
+                    var applicationProperties = Application.Current.Properties;
+                    applicationProperties["VLC_MediaPlayerElement_Position"] = MediaPlayer?.Position;
```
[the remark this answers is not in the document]
kakone: Damn, you're right, I did not think about this case. I'm not sure either that's the best way to go. Perhaps, we should let the developper manage the app lifecycle (and just provide practical methods to make it easier for him to recreate the videoview).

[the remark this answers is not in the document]
kakone: I often had more problems by defining the time rather than the position. That's why I always use the position. But, I have not retested for a long time, it would be interesting to know what is better.

mfkl:
> Did you intend to make the PR against 3.x instead of master?

Fixed

LibVLCSharp.Forms/Shared/Themes/Generic.xaml
kakone: The default icons value are for the Material Icons web font. In my mind, the font should not be redistributed. If the developper wants to use the default icon values, he must include the Material Icons font in his project (we can't do it automatically in a nuget package anyway). Otherwise, he can override the buttons styles to set its own icons (image files or icons font values).

LibVLCSharp.Forms/Shared/PlaybackControls.xaml.cs
```diff
@@ -0,0 +1,1384 @@
...
+
+        /// <summary>
+        /// Identifies the <see cref="ForeColor"/> dependency property.
+        /// </summary>
+        public static readonly BindableProperty ForeColorProperty = BindableProperty.Create(nameof(ForeColor), typeof(Color),
+            typeof(PlaybackControls));
+        /// <summary>
+        /// Gets or sets the button color.
+        /// </summary>
+        public Color ForeColor
```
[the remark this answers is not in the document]
kakone: OK to rename it to `Foreground`. I didn't want to use `TextColor` because it's mostly for the color of the icons.

LibVLCSharp.Forms/Shared/PlaybackControls.xaml.cs
```diff
@@ -0,0 +1,1384 @@
...
+        {
+            UpdateTime(MediaPlayer?.Position);
+        }
+
+        private void MediaPlayer_MediaChanged(object sender, MediaPlayerMediaChangedEventArgs e)
+        {
+            Reset();
+        }
+
+        private void MediaPlayer_NothingSpecial(object sender, EventArgs e)
```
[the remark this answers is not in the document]
kakone: I added it by security because I don't know when it can be called.

LibVLCSharp.Forms/Shared/Themes/Generic.xaml
jeremyVignelles: Will the code points be the same for all iconic fonts? Can we replace it easily with another font? Wouldn't it be better to have XAML vectorial images resources instead of characters?

LibVLCSharp.Forms.MediaPlayerElement/LibVLCSharp.Forms.MediaPlayerElement.csproj
```diff
@@ -0,0 +1,77 @@
+﻿<Project Sdk="MSBuild.Sdk.Extras/1.6.68">
```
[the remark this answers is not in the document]
mfkl: Done

Sample/Sample.Android/Assets/AboutAssets.txt
mfkl: removed

LibVLCSharp.Forms/Shared/PlaybackControls.xaml.cs
kakone: The UpdateTime method updates the **remaining** time label (the method should be renamed to UpdateRemainingTime). I can't show the remaining time without length and position (the time can't help in this case).
The UpdatePosition method is to update the seek bar position.
There are different methods because when the user moves the seek bar, I update the remaining time immediatly, and the stream position is not updated until the user has finished to move the seek bar.

LibVLCSharp.Forms/Shared/Themes/Generic.xaml
kakone: > Will the code points be the same for all iconic fonts? Can we replace it easily with another font?

You should override all buttons style to set the text property with the correct code and set the IconFontFamily property.

> Wouldn't it be better to have XAML vectorial images resources instead of characters?

I would have preferred to use vectorial images (SVG or XAML vectorial) but it's complicated in Xamarin Forms, you need some third party libraries. The better solution would be to have our own icons font (but I'm not a graphic designer :smiley: ).

mfkl: Font Awesome could be an interesting option as well https://github.com/matthewrdev/fa2cs

We should probably use the platform's icons when possible and fallback on custom ones for the missing icons (cast and close captions and likely more).

kakone: Or perhaps we can easily create an icon font from [these SVGs](https://code.videolan.org/videolan/vlc-android/tree/master/assets/images/svg).
I just tested with [IcoMoon app](https://icomoon.io/app/#/select), it seems easy to convert SVG to icons font. I need to test it with the MediaElement.

jeremyVignelles: I'm pretty sure that there are tools out there to convert svg to xaml. Isn't it possible to use <Path> in Xamarin.Forms? I didn't read which icons are required, but for play/pause, drawing rectangles/triangles should be enough

kakone: You can't use a Path element in Xamarin.Form Xaml like in WPF. You must use a third party library like [SkiaSharp ](https://github.com/mono/SkiaSharp) to do an equivalent thing.
Icons font is the better way to display vectorial icons in Xamarin.Forms.

kakone: I made a [custom icon font](https://github.com/kakone/LibVLCSharp.Forms.MediaPlayerElement/raw/master/Sample/Sample.Android/Assets/PlaybackControlsIcons.ttf) with the same identifiers than the Material Icons font (I used the [VLC Android SVGs](https://code.videolan.org/videolan/vlc-android/tree/master/assets/images/svg) and [IcoMoon app](https://icomoon.io/app/#/select) to generate the icon font).

So, we just need to replace the code by :
```
<OnPlatform x:Key="IconFontFamily" x:TypeArguments="x:String"
    Android = "PlaybackControlsIcons.ttf#PlaybackControlsIcons"
    iOS = "PlaybackControlsIcons" />
```

LibVLCSharp.Forms/Shared/PlaybackControls.xaml.cs
mfkl: Is that really necessary? It's fine for me, mostly public properties and short private handlers.

LibVLCSharp.Forms/Shared/Themes/Generic.xaml
mfkl: Done. I'm not sure we want those icons on iOS though.

mfkl: Can't get this .ttf to work on iOS and I won't spend more hours trying to.

Additionally, expecting and using icons in the framework part (libvlcsharp.forms) that are _required_ to be included by each client part (end user apps) does not seem like a smooth experience to me, unless we automatically fallback on defaults system icons (or text if no default).

LibVLCSharp.Forms/Shared/MediaPlayerElement.xaml
```diff
@@ -0,0 +1,27 @@
...
+                    <SwipeGestureRecognizer Swiped="GestureRecognized" />
+                </Grid.GestureRecognizers>
+
+                <Grid.RowDefinitions>
+                    <RowDefinition Height="*" />
+                    <RowDefinition Height="Auto" />
+                </Grid.RowDefinitions>
+
+                <ContentPresenter Grid.RowSpan="2" Content="{TemplateBinding VideoView}" />
+                <ContentPresenter Content="{TemplateBinding Content}" />
```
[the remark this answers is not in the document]
mfkl: Yep, looks fine until video is played.



Then when the video starts it paints all over the window (including on the playback controls).

LibVLCSharp.Forms/Shared/Themes/Generic.xaml
mfkl: VLC app icons are probably licensed GPLv2 (like the apps) so we probably cannot ship them with LibVLCSharp.Forms.

mfkl:
How can we make it really easy for users to hide *any* element/button from the playback controls that they might not want in their app (for example, chromecast, or zoom)?

LibVLCSharp.Forms/Shared/Themes/Generic.xaml
mfkl: Added the FontAwesome icon. The free pack seems open enough. And used this https://github.com/matthewrdev/fa2cs because the codes were giving me a headache.
Briefly looked into packaging those into the end user binary (apk, .app, whatever) from LibVLCSharp.Forms and:
- it doesn't look straightforward though still likely possible with a custom target
- it doesn't seem to be the convention for libraries to do that, though I thought it'd be a smoother dev experience (just like adding a WPF MediaElement and be done with it).
Thoughts?

kakone:
You have the IsCastButtonVisible and IsZoomButtonVisible properties on the PlaybackControls class :
```csharp
<vlc:MediaPlayerElement MediaPlayer="{Binding MediaPlayer}" LibVLC="{Binding LibVLC}">
    <vlc:MediaPlayerElement.PlaybackControls>
        <vlc:PlaybackControls IsCastButtonVisible="False" IsZoomButtonVisible="False" />
    </vlc:MediaPlayerElement.PlaybackControls>
</vlc:MediaPlayerElement>
```

mfkl:
Indeed (missed that)

mfkl:
Any reason you used the `Scale` property instead of the `AspectRatio` one for the Zoom?

LibVLCSharp.Forms/Shared/Themes/Generic.xaml
kakone: In my idea, it is not a smooth dev experience, but I thought to ask the dev to copy the font manually. Because if he wants to use another font or images, he doesn't need this FontAwesome.ttf file in his final package.

mfkl: Yes, but
- most people won't bother and will use the default icons (like the usual mediaElement)
- the font we provide could be disable with a csproj boolean property `<UseFontAwesome>false</UseFontAwesome>` in the user project.

kakone:
> Any reason you used the `Scale` property instead of the `AspectRatio` one for the Zoom?

No, I don't see why I didn't use `AspectRatio`.

LibVLCSharp.Forms/Shared/MediaPlayerElement.xaml.cs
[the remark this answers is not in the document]
mfkl: > Perhaps, we should let the developper manage the app lifecycle (and just provide practical methods to make it easier for him to recreate the videoview).

Yes, maybe. There seems to be no way for us (from the library code) to know when the user's app is going to background. 
The less adjustment the user has to make to his project to get up and running the better. The solution you have for now works well for most cases. 

For more advanced cases (multiple videos, background audio, etc), users will want to handle the lifecycle/playback themselves, and they can (by not firing those messages). I guess we could make those message callbacks public..?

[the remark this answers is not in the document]
mfkl: Let's try it as it is for now and wait for feedback.

LibVLCSharp.Forms/Shared/Themes/Generic.xaml
mfkl: Think I found a more gracious way.
- Bundle fonts with `EmbeddedResource` with LibVLCSharp.Forms (if not told otherwise by user's csproj).
- Use XF platform effects to apply them.
Started work here https://code.videolan.org/mfkl/LibVLCSharp/commit/3058d14358c2f6009b7cf2ec23acd5f74d92da92 but will finish it in a next update.

LibVLCSharp.Forms/Shared/MediaPlayerElement.xaml.cs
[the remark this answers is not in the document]
jeremyVignelles: @mfkl : Do you suggest something like this?

```cs
var state = mediaElement.SaveState();
// Serialize the state the way you want

// On restore
mediaElement.RestoreState(state);
```
Do you have something else in mind?

LibVLCSharp.Forms/Shared/MediaPlayerElement.xaml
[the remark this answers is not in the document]
jeremyVignelles: What was decided here?

LibVLCSharp.Forms/Shared/PlaybackControls.xaml.cs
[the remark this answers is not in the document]
jeremyVignelles: This is an event handler, we can safely remove that (nothing will be called if no event handler, because we call it as `?.Invoke...`)

@mfkl : Do we really need to expose this event in the core?

LibVLCSharp.Forms/Shared/PlaybackControls.xaml.cs
jeremyVignelles: things are getting a bit complicated starting with `OnApplyTemplate` at line 602. I really think that this component is too big to be maintained efficiently, and it only has 3 buttons yet...

LibVLCSharp.Forms/Shared/PlaybackControls.xaml.cs
jeremyVignelles: I think there are some big confusions here. `position` is used interchangeably with `time`.
Let's agree on naming:
- `position` is the position between 0 (begin) and 1 (end). It is a floating point value.
- `time` is the number of milliseconds elapsed since the beginning of the playback. It is a `long` and stored between 0 and `length`

line 688 : It is called with a `Position`. Chances are that with the sample we chose, the Length only changes when `Position == 0`, but this will not always be the case.
line 1295 : A `TotalMilliseconds` is passed
line 1207 : null is implicitely passed.

Then, below in this method, if `position` is not null, we assign it to `time` and is interpreted as a time.

I think we should also add comments on top of this method. I can see that the given position is nullable. How is `null` interpreted?

I think that `position` should be renamed to `time` and be set to a `long` value

LibVLCSharp.Forms/Shared/PlaybackControls.xaml.cs
```diff
@@ -0,0 +1,1369 @@
...
+                return;
+            }
+
+            try
+            {
+                var mediaPlayer = MediaPlayer;
+                var state = mediaPlayer?.State;
+                var length = mediaPlayer == null || state == VLCState.Ended || state == VLCState.Error || state == VLCState.NothingSpecial ||
+                    state == VLCState.Stopped ? 0 : mediaPlayer.Length;
+                var time = position == null ? (SeekBar.Value * length / SeekBar.Maximum) : (double)position;
```
jeremyVignelles: This is equivalent to:

```suggestion
                var time = position ?? (SeekBar.Value * length / SeekBar.Maximum);
```

jeremyVignelles:
Hi,
I have tested the sample application on my android phone, and I have a few UX concerns:

- Seeking doesn't feel smooth (and it surely isn't due to the timer). When on computer, I'm used to drag the slider too look for a specific position in the video. This implementation explicitely disallows this behavior. Any thoughts?
- What's the point of the "zoom" button? It does zoom, but when I'm in portrait mode, the view it gives seems useless to me.
- When casting, I don't like the spinning animation. I'd rather have an overlay view with an "indeterminate progressbar" while it's scanning, and having results appearing as they are available.
- When casting, I've been a bit confused by the empty dialog that I got. We should probably show a message like "Scan completed, but no devices found".

Then, about the controls customization, I'm not sure that being able to add/remove buttons will be enough to cover most use cases. Can we make it more "composable"? Let me explain.

The current implementation of the PlaybackControls is too monolithic. You can customize what is displayed by overriding the template, but when new features will come (speed control, snapshot, subtitles...) it will be a nightmare to maintain (1300 lines already).
What about splitting the monolith into one control per button? Currently, we would need a `PlayPauseButton`, a `ZoomButton`, a `CastButton` and a `VideoPositionSlider`. Each of these pieces only rely on the MediaPlayer to provide their features, and talk directly with it.

Each feature becomes easier to implement, and composition becomes easier too : It's just a matter of putting predefined controls inside a template.

What do you think?

LibVLCSharp.Forms/Shared/MediaPlayerElement.xaml.cs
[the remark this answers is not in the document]
mfkl: Something like that. 

If we store the `MediaElementState` in the `Application Properties` we can make the parameter of `RestoreState` optional, so that the user does not have hold of it. But essentially yes.

LibVLCSharp.Forms/Shared/PlaybackControls.xaml.cs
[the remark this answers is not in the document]
mfkl: I mean it's not a dead event, it can be raised from libvlc. 

How relevant it is I guess that's debatable... but we can leave for completeness for now. Can reconsider for 4.0 I guess.

mfkl:
> This implementation explicitly disallows this behavior

What do you mean?

> What's the point of the "zoom" button? It does zoom, but when I'm in portrait mode, the view it gives seems useless to me.

Started a refactoring to align the aspect ratio behavior with what VLC desktop does. Not finished yet:
https://code.videolan.org/mfkl/LibVLCSharp/commit/5cd0bbd7f0d8803803f98ec826896d199ebd256a, though I don't think this is blocking, we can update this soon.

Then we could do a zoom feature ^^

> When casting, I don't like the spinning animation.

I guess that's just personal taste :-) No strong feeling about this, but this is also not blocking and can be updated later.

> When casting, I've been a bit confused by the empty dialog that I got. We should probably show a message like "Scan completed, but no devices found".

Yes.

> The current implementation of the PlaybackControls is too monolithic. You can customize what is displayed by overriding the template, but when new features will come (speed control, snapshot, subtitles...) it will be a nightmare to maintain (1300 lines already).

But you can ignore half of those. Put the bindable properties in a region and (mostly) forget about it.

> What about splitting the monolith into one control per button? Currently, we would need a PlayPauseButton, a ZoomButton, a CastButton and a VideoPositionSlider. Each of these pieces only rely on the MediaPlayer to provide their features, and talk directly with it.

> What do you think?

I'm not against the idea, but I also don't think a lot of one line event handlers / properties add much maintenance complexity, if well put together.

LibVLCSharp.Forms/Shared/MediaPlayerElement.xaml.cs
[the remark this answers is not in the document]
jeremyVignelles: What about saving the state of multiple players? Wouldn't that cause conflicts?

jeremyVignelles:
> > This implementation explicitly disallows this behavior
> 
> What do you mean?

I mean that the timer rate-limits the seek. each time the slider is moved, the timer is reset and waits to see if another event is sent. If I move my slider continuously, nothin happens until I release the slider and wait for the timer to elapse.

> > What's the point of the "zoom" button? It does zoom, but when I'm in portrait mode, the view it gives seems useless to me.
> 
> Started a refactoring to align the aspect ratio behavior with what VLC desktop does. Not finished yet:
> https://code.videolan.org/mfkl/LibVLCSharp/commit/5cd0bbd7f0d8803803f98ec826896d199ebd256a, though I don't think this is blocking, we can update this soon.
> 
> Then we could do a zoom feature ^^
OK

> > When casting, I don't like the spinning animation.
> 
> I guess that's just personal taste :-) No strong feeling about this, but this is also not blocking and can be updated later.

Well, OK. Let's talk about that later

> I'm not against the idea, but I also don't think a lot of one line event handlers / properties add much maintenance complexity, if well put together.

If it was only event handlers that forwards things with a similar structure, like the event manager, I wouldn't mind too much because it's mostly duplicates that you can read diagonally.
However, here, each event handler has its own action, and all the control features are mixed into a single file.
The buttons handlers are not one-liners either:
- The Play/pause button needs to change based on the player status
- The zoom feature as implemented like VLC for android, will cycle between several states
- The cast needs to display a popup
- The slider needs to react to position changed events...

I'm aware that it adds a lot of files, but when you have a bug in, say, the seek bar, you know where to look at

LibVLCSharp.Forms/Shared/MediaPlayerElement.xaml.cs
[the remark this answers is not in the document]
jeremyVignelles: I'm not sure we should handle lifecycle events here. The user has the MediaPlayer already, and he can save/restore the state programmatically, as in any other LVS application.

mfkl:
I meant the first 570 lines

LibVLCSharp.Forms/Shared/Themes/Generic.xaml
mfkl: Turns out we will be able to use VLC iOS/Android icons after all...

kakone:
> I mean that the timer rate-limits the seek. each time the slider is moved, the timer is reset and waits to see if another event is sent. If I move my slider continuously, nothin happens until I release the slider and wait for the timer to elapse.

Yes, it is the desired behavior. In fact, seeking is very slow with internet streams due to buffering and I had a lot of problems when I did multiple seeking operations in a short lap time. So I prefered to introduce this timer to not have these problems.

LibVLCSharp.Forms/Shared/MediaPlayerElement.xaml.cs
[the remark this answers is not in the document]
mfkl: > What about saving the state of multiple players? Wouldn't that cause conflicts?

We could use 
```
var positionKey = $"VLC_{MediaPlayer.NativeReference}_MediaPlayerElement_Position";
```
since the MediaPlayer is not disposed when the app is backgrounded. If the user wants to dispose it, then they just don't fire the lifecycle message and handle everything themselves.

> I'm not sure we should handle lifecycle events here.

We just offer to do the basic lifecycle management by offering the user to send a simple message. They are still free to _not_ send the message and handle everything themselves, but we still handle the basic use case gracefully.

kakone:
> The current implementation of the PlaybackControls is too monolithic. You can customize what is displayed by overriding the template, but when new features will come (speed control, snapshot, subtitles...) it will be a nightmare to maintain (1300 lines already).

In this case, you can use the Style properties to customize the buttons (CastButtonStyle, ZoomButtonStyle, ...), or use the ButtonColor property to modify the color. You have also some IsVisible properties to show/hide the buttons (IsCastButtonVisible, ...).
And you can also add some custom controls in the ButtonBarStartArea or ButtonBarEndArea properties.
How to make it more customizable?

mfkl:
> Yes, it is the desired behavior. In fact, seeking is very slow with internet streams due to buffering and I had a lot of problems when I did multiple seeking operations in a short lap time. So I prefered to introduce this timer to not have these problems.

I understand, though I just tried on desktop and Android VLC apps, in paused mode, and when moving the slide the frame still updates (at least without issue for local files). 

> When casting, I don't like the spinning animation.

Just checked and it appears the Android app displays the chromecasts immediately. Haven't checked the code but I assume it checks for chromecasts immediately when the app starts, not when the user taps the cast button (so, no/little waiting).

I guess my point is that we have to decide now (or soon) if we want to align with current VLC apps behavior (assuming they are the same across platforms) or if we want to go our own way. 
(Or align by default and provide some determined amount of limited customization to the dev).

mfkl:
> Yes, it is the desired behavior. In fact, seeking is very slow with internet streams due to buffering and I had a lot of problems when I did multiple seeking operations in a short lap time. So I prefered to introduce this timer to not have these problems.

I removed the timer code to align with VLC. 

We still might add it back later as an optional feature, because it may indeed make sense for some users. We may also use a default value for the network caching param (like VLC Android [does](https://github.com/videolan/vlc-android/blob/3d3a9e050081ce45991e475d54cdfc592aadb9ab/libvlc/src/org/videolan/libvlc/Media.java#L782)).

jeremyVignelles:
Can you really set network caching defaults? the MediaPlayer is passed from the outside, right?

mfkl:
> Just checked and it appears the Android app displays the chromecasts immediately. Haven't checked the code but I assume it checks for chromecasts immediately when the app starts, not when the user taps the cast button (so, no/little waiting).

Default behavior:
- When MediaElement is created, look for chromecasts automatically. 
- If found, show cast button.
- If cast button pressed and only 1 chromecast found, connect to it automatically, else show a popup to choose.

Customization:
- Allow not to look for chromecasts upon MediaElement creation (extra bindable property on MediaElement I guess).
- Hide cast button (done already).

Thoughts?

jeremyVignelles:
I'm not sure if it is a Bindable property or a regular property. Does that make sense to enable it later?
The behavior you're describing seems consistent with VLC for Android. At first, I was a bit concerned about the fact that we can't debug why the button is hidden, but the developer can still listen to the same events and show logs appropriately.

Rather than hidden, can we make the button disabled? I'm not a huge fan of UI that jumps away when you want to click on it...

mfkl:
> Rather than hidden, can we make the button disabled? I'm not a huge fan of UI that jumps away when you want to click on it...

Here again I guess it's personal taste :-P  but that's how VLC apps behave on all platforms AFAIK.

kakone:
> When MediaElement is created, look for chromecasts automatically.

You must also update the cast button when a chromecast is added (or removed) to the network after the MediaElement creation.

mfkl:
I'm gonna merge this as-is for a pre-release build because I think it'd be nice to have early user feedback.

Created 3 issues to follow up on:
- [Rework Chromecast behavior](https://code.videolan.org/videolan/LibVLCSharp/issues/179)
- [Icon management](https://code.videolan.org/videolan/LibVLCSharp/issues/178)
- [Refactor and complete the AspectRatio feature](https://code.videolan.org/videolan/LibVLCSharp/issues/177)

Feel free to let me know if I missed anything so far. 

We can also investigate how to better organize the code (though not a big fan of partials).

mfkl:
Closing this for now. pre-release build is at https://www.nuget.org/packages/LibVLCSharp.Forms/3.1.1-alpha.

Feel free to open new PRs for issues linked above ^. Will merge this branch in 3.x and do a stable release in a bit.

PR #53, opened 2019-06-13
Extend Media Configuration API

mfkl:
Extend the media configuration helper:
- Works on both MediaPlayer and Media
- Add HW configuration support for more platforms (can be called from shared code)
Tested:
- [x] Android (existing) 
- [x] iOS
- [x] Windows
- [x] macOS (should be the same as iOS)
- Add network and file caching options

https://code.videolan.org/videolan/LibVLCSharp/issues/132

Will squash.

### DIFF
--- a/LibVLCSharp/Shared/Media.cs
+++ b/LibVLCSharp/Shared/Media.cs
@@ -309,8 +309,7 @@ namespace LibVLCSharp.Shared
 
             foreach(var option in mediaConfiguration.Build())
             {
-                if(!string.IsNullOrWhiteSpace(option))
-                    AddOption(option);
+                AddOption(option);
             }
         }
 
--- a/LibVLCSharp/Shared/Media.cs
+++ b/LibVLCSharp/Shared/Media.cs
@@ -307,7 +307,11 @@ namespace LibVLCSharp.Shared
         {
             if (mediaConfiguration == null) throw new ArgumentNullException(nameof(mediaConfiguration));
 
-            AddOption(mediaConfiguration.Build());
+            foreach(var option in mediaConfiguration.Build())
+            {
+                if(!string.IsNullOrWhiteSpace(option))
+                    AddOption(option);
+            }
         }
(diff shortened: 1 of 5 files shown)


LibVLCSharp/Shared/MediaConfiguration.cs
```diff
@@ -0,0 +1,110 @@
...
+#endif
+            }
+
+        }
+
+        /// <summary>
+        /// Builds the current MediaConfiguration for consumption by libvlc (or storage)
+        /// </summary>
+        /// <returns></returns>
+        public string[] Build() => _options.Values.ToArray();
```
jeremyVignelles: Should we add `.Where(opt => !string.IsNullOrEmpty(opt))` here to filter out empty options, and remove the check from the caller?

mfkl: > remove the check from the caller?

Which check are you referring to exactly?

LibVLCSharp/Shared/Media.cs
```diff
@@ -307,7 +307,11 @@ public void AddOption(MediaConfiguration mediaConfiguration)
         {
             if (mediaConfiguration == null) throw new ArgumentNullException(nameof(mediaConfiguration));
 
-            AddOption(mediaConfiguration.Build());
+            foreach(var option in mediaConfiguration.Build())
+            {
+                if(!string.IsNullOrWhiteSpace(option))
```
jeremyVignelles: To be removed if other comment is applied in Build()

LibVLCSharp/Shared/MediaConfiguration.cs
jeremyVignelles: see other comment

mfkl:
Merged https://code.videolan.org/videolan/LibVLCSharp/commit/6995b9f8692ffd0215c97ed9907aeb0beb50765a

PR #54, opened 2019-06-14
Add VideoView's DataContext passthrough to Foreground window

Tigermeet:
This allows video overlaying elements inside XAML \<VideoView\> tag to be data bound to viewmodel's properties in MVVM scenarios.


mfkl:
Thanks!

jeremyVignelles:
After thinking about it a little, I'm starting to wonder... Wouldn't it be better if we handled the DataContext binding from the ForegroundWindow class? I mean, we probably have everything needed there (we pass the WindowsFormsHost, from which we get the DataContext), and there is another catch: 

What if the DataContext is changed in the background element? I think we should listen to DataContext changes events.

mfkl:
Good point.

PR #56, opened 2019-06-18
Refactored WPF DataContext passthrough

jeremyVignelles:
We now subscribe to the DataContextChanged event to make sure the
DataContext is always in sync with the outside

This is a follow-up of the discussion in #54

### DIFF
--- a/LibVLCSharp.WPF/ForegroundWindow.xaml.cs
+++ b/LibVLCSharp.WPF/ForegroundWindow.xaml.cs
@@ -30,9 +30,17 @@ namespace LibVLCSharp.WPF
         {
             InitializeComponent();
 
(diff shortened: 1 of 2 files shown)


Tigermeet:
Agreed, this is much better. Great work!

mfkl:
remove the `this.` please, you know I don't like this convention ;-)

LGTM otherwise.

jeremyVignelles:
Please add a tool that fails build when I don't follow your conventions 😄 Switching between conventions is not that easy

mfkl:
I still haven't managed to make editorconfig work locally, with msbuild.sdk.extras.

mfkl:
Merged, thanks!

PR #60, opened 2019-07-17
Chromecast refactoring

mfkl:
- [x] When MediaElement is created, look for chromecasts immediately.
- [x] If found, show cast button.
- [x] If cast button pressed and only 1 chromecast found, connect to it automatically, else show a popup to choose.
- [x] Update the cast button when a chromecast is added (or removed) to the network after the MediaElement creation.
- [x] Allows disconnecting from a chromecast (including https://github.com/videolan/libvlcsharp/commit/f76372ca729d5b4658fafcdfd5f0af6e46deffa0)
- [x] Works on all platforms.
- [x] Allow not to look for chromecasts upon MediaElement creation (extra bindable property on MediaElement I guess).
```
<vlc:MediaPlayerElement 
     MediaPlayer="{Binding MediaPlayer}" 
     LibVLC="{Binding LibVLC}" 
     EnableRendererDiscovery="False" />
```
- [ ] See if we can move some of this code out.

I have not tested what happens when a chromecast is disconnected while we are casting to it. Probably should update some stuff on our side.


mfkl:
Hmm, unsure if `EnableRendererDiscovery` should be on the `MediaPlayerElement` or on `PlaybackControls`.

mfkl:
Merged

PR #63, opened 2019-09-04
WPF - ALT+F4 issue

srollinet:
### Description of Change ###

Move the focus to the parent window when ALT+F4 is pressed to avoid closing only the Foreground window.

### Issues Resolved ### 

None

### API Changes ###

 None

### Platforms Affected ### 

- WPF

When the user hits ALT+F4, it closes the background window instead of the foreground window.

Not applicable

### Testing Procedure ###

With the WPF sample, (Example 1 and Example 2):
- click on the play button -> media starts playing
- hit Alt+f4 -> the background window is closed, not only the foreground window

### PR Checklist ###

### DIFF
--- a/LibVLCSharp.WPF/ForegroundWindow.xaml.cs
+++ b/LibVLCSharp.WPF/ForegroundWindow.xaml.cs
@@ -2,6 +2,7 @@
 using System;
 using System.Diagnostics;
 using System.Windows;
+using System.Windows.Input;
 
 namespace LibVLCSharp.WPF
 {
@@ -107,6 +108,14 @@ namespace LibVLCSharp.WPF
         void Wndhost_Closing(object sender, System.ComponentModel.CancelEventArgs e)
         {
             Close();
+        }
+
+        protected override void OnKeyDown(KeyEventArgs e)
+        {
+            if (e.Key == Key.System && e.SystemKey == Key.F4)
+            {
+                _wndhost.Focus();
+            }
         }
     }
 }


mfkl:
Hi @srollinet,

I'm not clear on what the actual problem is. 

> it closes the background window instead of the foreground window.

How do you observe this exactly?

To me, it'd appear that the focus issue is only in the sample code. When clicking play, the button gets the focus instead of remaining on the window.

Wouldn't something like this be enough?
```
if (!VideoView.MediaPlayer.IsPlaying)
{
    VideoView.MediaPlayer.Play(new Media(_libVLC,
    "https://download.blender.org/peach/bigbuckbunny_movies/BigBuckBunny_320x180.mp4", FromType.FromLocation));
+   Focus();
}
```

srollinet:
The problem with the transparent foreground windows, is that if you click on it (or any of its children), the transparent window takes the focus.

Then, if you hit Alt+F4, it closes the focused windows -> the transparent one.
So, the idea, if the transparent window receives an Alt+F4 KeyPressed event, the focus is transfered on the parent windows and both windows are closed.

mfkl:
> So, the idea, if the transparent window receives an Alt+F4 KeyPressed event, is to transfer the focus to the parent windows and both windows are closed.

Is there a case where we'd want the transparent window to ever get the focus?

I'm wondering whether a more generic solution to your problem would be a good idea. What if the user wants to move the window around with Windows Key + arrows?

srollinet:
> Is there a case where we'd want the transparent window to ever get the focus?

The window itself, probably not, but  its content could be.

> I'm wondering whether a more generic solution to your problem would be a good idea. What if the user wants to move the window around with Windows Key + arrows?

Totally agree! Now that you mention it, I remember that last time I pressed Win + Arrow, it didn't work, but I did not really pay attention in the moment. Now I realize it probably moved the transparent window 😑 

I tried something like that

     protected override void OnGotKeyboardFocus(KeyboardFocusChangedEventArgs e) => Keyboard.Focus(_wndhost);

But then, with the sample code, the buttons are clickable only once

Also tried to set `Focusable="False"` on the window, but it doesn't change anything.

jeremyVignelles:
What options do we have?
- Keep as is : Focus works for items in the overlay, but shortcuts like ALT+F4, WIN+arrows might not work.
- Disallow all focus on the overlay : Prevents things from acting normally, does not seem good.
- Handle case-by-case the events like Alt+F4 and WIN shortcuts. Might break things in some cases.

I think I hate the third option less... When the Windows key is pressed, I'd say that it can be possible to focus the window behind, but I don't know how tab order would work then...

Things might break in subtle ways, but on the other hand, I don't want it to stay like that...

mfkl:
Tried a couple more solutions, but I find them too intrusive and it will likely cause problems.

Let's go with the ALT+F4 exception for now (and maybe the winkey + arrows). It's good enough for now, and users are warned of airspace issues anyway (https://code.videolan.org/videolan/LibVLCSharp/blob/master/LibVLCSharp.WPF/README.md#airspace-limitations)

jeremyVignelles:
I'm OK with the code, but indeed we need to handle the win key as well. Will that be in that PR or in another one?

mfkl:
I say we take those changes in first, and then see which other shortcuts would really make sense to handle? 

Might be useful in the future to allow users to enable/disable that input handling by us.

jeremyVignelles:
LGTM, let's make another PR for other use cases.

mfkl:
Merged https://github.com/videolan/libvlcsharp/commit/0002a659360b64b9d5f7834abd7c2ecd0350ce0d, thanks!

PR #66, opened 2019-09-17
Fix libvlc logs bugs

mfkl:
### Description of Change ###

- Fix logging on all instances bug.
- Fix linux crash by allocating dynamically buffer size.

### Issues Resolved ### 

- fixes https://code.videolan.org/videolan/LibVLCSharp/issues/213
- fixes https://code.videolan.org/videolan/LibVLCSharp/issues/216

### API Changes ###

 None

### Platforms Affected ### 

- Core (all platforms)

### Testing Procedure ###

Try enabling/disabling logging on all supported platforms and devices and see if it works fine.

- [x] Windows
- [x] Android
- [ ] iOS
- [ ] macOS
- [x] Linux

### PR Checklist ###

### DIFF
--- a/LibVLCSharp/Shared/Helpers/MarshalUtils.cs
+++ b/LibVLCSharp/Shared/Helpers/MarshalUtils.cs
@@ -64,7 +64,90 @@ namespace LibVLCSharp.Shared.Helpers
             public static extern int vsnprintf_windows(IntPtr buffer, UIntPtr size, IntPtr format, IntPtr args);
         }
 
-        internal static int vsnprintf(IntPtr buffer, UIntPtr size, IntPtr format, IntPtr args)
+        #region logging
+
+        internal static string GetLogMessage(IntPtr format, IntPtr args)
+        {
+            // special marshalling is needed on Linux desktop 64 bits.
+            if (PlatformHelper.IsLinuxDesktop && PlatformHelper.IsX64BitProcess)
+            {
+                return LinuxX64Callback(format, args);
+            }
+
+            var byteLength = vsnprintf(IntPtr.Zero, UIntPtr.Zero, format, args) + 1;
+            if (byteLength <= 1)
+                return string.Empty;
+
+            var buffer = IntPtr.Zero;
+            try
+            {
+                buffer = Marshal.AllocHGlobal(byteLength);
+                vsprintf(buffer, format, args);
+                return buffer.FromUtf8();
+            }
+            finally
+            {
+                Marshal.FreeHGlobal(buffer);
+            }
+        }
+
+        static string LinuxX64Callback(IntPtr format, IntPtr args)
+        {
+            // The args pointer cannot be reused between two calls. We need to make a copy of the underlying structure.
+            var listStructure = PtrToStructure<VaListLinuxX64>(args);
+            var byteLength = 0;
+            UseStructurePointer(listStructure, listPointer =>
+            {
+                byteLength = Native.vsnprintf_linux(IntPtr.Zero, UIntPtr.Zero, format, listPointer) + 1;
+            });
+
+            var utf8Buffer = IntPtr.Zero;
+            try
+            {
+                utf8Buffer = Marshal.AllocHGlobal(byteLength);
+
+                return UseStructurePointer(listStructure, listPointer =>
+                {
+                    Native.vsprintf_linux(utf8Buffer, format, listPointer);
+                    return utf8Buffer.FromUtf8();
+                });
+            }
+            finally
+            {
+                Marshal.FreeHGlobal(utf8Buffer);
+            }
+        }
+
+        static string UseStructurePointer<T>(T structure, Func<IntPtr, string> action)
+        {
+            var listPointer = IntPtr.Zero;
+            try
+            {
+                listPointer = Marshal.AllocHGlobal(Marshal.SizeOf(structure));
+                Marshal.StructureToPtr(structure, listPointer, false);
+                return action(listPointer);
+            }
+            finally
+            {
+                Marshal.FreeHGlobal(listPointer);
+            }
+        }
+
+        static void UseStructurePointer<T>(T structure, Action<IntPtr> action)
+        {
+            var listPointer = Marshal.AllocHGlobal(Marshal.SizeOf(structure));
+            try
+            {
+                Marshal.StructureToPtr(structure, listPointer, false);
+                action(listPointer);
+            }
+            finally
+            {
+                Marshal.FreeHGlobal(listPointer);
+            }
+        }
+
+        static int vsnprintf(IntPtr buffer, UIntPtr size, IntPtr format, IntPtr args)
         {
 #if ANDROID
             return Native.vsnprintf_linux(buffer, size, format, args);
@@ -81,7 +164,7 @@ namespace LibVLCSharp.Shared.Helpers
 #endif
         }
 
-        internal static int vsprintf(IntPtr buffer, IntPtr format, IntPtr args)
+        static int vsprintf(IntPtr buffer, IntPtr format, IntPtr args)
         {
 #if ANDROID
             return Native.vsprintf_linux(buffer, format, args);
@@ -97,6 +180,8 @@ namespace LibVLCSharp.Shared.Helpers
             return -1;
 #endif
         }
+        
+        #endregion
 
         /// <summary>
         /// Helper for libvlc_new
@@ -538,6 +623,15 @@ namespace LibVLCSharp.Shared.Helpers
         {
             foreach (var ptr in ptrs)
                 Marshal.FreeHGlobal(ptr);
+        }
+
+        [StructLayout(LayoutKind.Sequential, Pack = 4)]
+        struct VaListLinuxX64
+        {
+            uint gp_offset;
+            uint fp_offset;
+            IntPtr overflow_arg_area;
+            IntPtr reg_save_area;
         }
     }
 
--- a/LibVLCSharp/Shared/Helpers/MarshalUtils.cs
+++ b/LibVLCSharp/Shared/Helpers/MarshalUtils.cs
@@ -120,30 +120,30 @@ namespace LibVLCSharp.Shared.Helpers
 
         static string UseStructurePointer<T>(T structure, Func<IntPtr, string> action)
         {
-            var listPointer = IntPtr.Zero;
+            var structurePointer = IntPtr.Zero;
             try
             {
-                listPointer = Marshal.AllocHGlobal(Marshal.SizeOf(structure));
-                Marshal.StructureToPtr(structure, listPointer, false);
-                return action(listPointer);
+                structurePointer = Marshal.AllocHGlobal(Marshal.SizeOf(structure));
+                Marshal.StructureToPtr(structure, structurePointer, false);
+                return action(structurePointer);
             }
             finally
             {
-                Marshal.FreeHGlobal(listPointer);
+                Marshal.FreeHGlobal(structurePointer);
             }
         }
 
         static void UseStructurePointer<T>(T structure, Action<IntPtr> action)
         {
-            var listPointer = Marshal.AllocHGlobal(Marshal.SizeOf(structure));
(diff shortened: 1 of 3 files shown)


mfkl:
Could you please try a build of that branch on linux @srollinet?

srollinet:
Tried with this code

    class Program
    {
        static void Main(string[] args)
        {
            Core.Initialize();

            var libVLC1 = new LibVLC();
            libVLC1.Log += (sender, eventArgs) => Console.WriteLine($"libVLC1 : {eventArgs.Message}");

            var libVLC2 = new LibVLC();
            libVLC2.Log += (sender, eventArgs) => Console.WriteLine($"libVLC2 : {eventArgs.Message}");
            var media = new Media(libVLC2, "/home/sro/Downloads/jellyfish-3-mbps-hd-h264.mkv", FromType.FromPath);
            var mp = new MediaPlayer(media);
            mp.Play();

            Console.ReadKey();
        }
    }

It crashes here: `var mp = new MediaPlayer(media);`

output is:

> libVLC1 : VLC media player - �"��
> libVLC1 : revision �"��
> libVLC2 : revision H���
> libVLC1 : VLC media player - %s
> libVLC2 : VLC media player - H���
> libVLC1 : configured with revision %s
> libVLC2 : VLC media player - %s
> libVLC2 : configured with revision %s
> 
> Process finished with exit code 139.

It doesn't crash if I subscribe only to libVLC1 events

srollinet:
It crashes here : https://github.com/videolan/libvlcsharp/pull/66/files#diff-453df8d527f3d7907bf7a51d991fb9aaR651

last log before the crash is "creating audio output"

LibVLCSharp/Shared/LibVLC.cs
```diff
@@ -403,6 +409,10 @@ void UnsetLog()
                     {
                         UnsetLog();
                     }
+                    else
```
jeremyVignelles: Why do you need to reset logs?

mfkl: Since the `EventHandler<LogEventArgs>` is given to the native side, it has to be updated on every sub/unsub otherwise when invoked it may not be the last version.

LibVLCSharp/Shared/LibVLC.cs
```diff
@@ -623,23 +633,30 @@ static void UpdateProgress(IntPtr data, IntPtr dialogId, float position, string
...
-            IntPtr buffer = IntPtr.Zero;
+            var buffer = IntPtr.Zero;
             try
             {
-                buffer = Marshal.AllocHGlobal(2048 + 1);
+                var byteLength = MarshalUtils.vsnprintf(IntPtr.Zero, UIntPtr.Zero, format, args) + 1;
+                if (byteLength <= 1) return;
+
+                buffer = Marshal.AllocHGlobal(byteLength);
                 var count = MarshalUtils.vsprintf(buffer, format, args);
```
jeremyVignelles: I would try to always use vsnprintf, but probably more a matter of taste.

LibVLCSharp/Shared/LibVLC.cs
jeremyVignelles: Can we pass the LibVLC instance then? That way, it doesn't change.

jeremyVignelles:
I'm afraid that @srollinet faced the true meaning of this line in the vsnprintf documentation:

> Internally, the function retrieves arguments from the list identified by arg as if va_arg was used on it, and thus the state of arg is likely to be altered by the call.

http://www.cplusplus.com/reference/cstdio/vsnprintf/

I'm afraid we can't do it that way, because C# doesn't have a proper va_list marshalling/va_copy system...

LibVLCSharp/Shared/LibVLC.cs
mfkl: What do you mean? We need to retrieve the logger in a static context from native code to raise events. How do we that with only a `LibVLC` instance?

mfkl:
Tested https://github.com/videolan/libvlcsharp/pull/66/commits/331c88b3c81bdcd6629b791dc5a77b401bafa90a on Windows, Android and WSL x64.

`linux-arm` (raspberry) will probably be an issue but we will see later when we get there.

Need access to a mac to test Apple platforms.

LibVLCSharp/Shared/LibVLC.cs
jeremyVignelles: Pass the libvlc instance (aka `this`) to the GCHandle.
Then, from the callback :
`libvlc._log...`

LibVLCSharp/Shared/LibVLC.cs
```diff
@@ -623,28 +633,27 @@ static void UpdateProgress(IntPtr data, IntPtr dialogId, float position, string
...
-                if (count > -1)
-                {
-                    var message = buffer.FromUtf8();
-                    GetLogContext(ctx, out var module, out var file, out var line);
+                var message = MarshalUtils.GetLogMessage(format, args);
+
+                GetLogContext(ctx, out var module, out var file, out var line);
 #if NET40
-                    Task.Factory.StartNew(() => _log?.Invoke(null, new LogEventArgs(level, message, module, file, line)));
+                Task.Factory.StartNew(() => logger?.Invoke(null, new LogEventArgs(level, message, module, file, line)));
```
jeremyVignelles: shall we lock the calls to the logger? (more a self reminder than a real comment)

mfkl: Why? The null-conditional check is thread-safe.

jeremyVignelles: Is the invocation atomic?
I see a risk of messing things up /deadlocking if someone unregister in the log callback. We should think carefully about this I guess.

mfkl: > someone unregister in the log callback. 

Yeah well, that's beyond asking for trouble lool

jeremyVignelles: Don't underestimate what users can do 😃. My point is that : What if you unregister a log callback while it is being invoked, say because your log callback takes some time to execute and that you unfortunately unregister the callback at the same time?

My rule of thumb is that if you lock() somewhere, then you should lock everywhere it's being used

srollinet:
I confirm 331c88b3c81bdcd6629b791dc5a77b401bafa90a works for me as well

LibVLCSharp/Shared/LibVLC.cs
mfkl: What you're saying would apply for all events and so far no bug report about this. And a lock in this case would really slow things down as it can be called quite a lot. 

I say let's leave it as it is for now, and see.

mfkl:
GCHandle is a value type, cannot be null, you're talking about `libvlc`?

jeremyVignelles:
GitHub lost my context, I was reviewing the last commit about `libvlc?.`

LibVLCSharp/Shared/Helpers/MarshalUtils.cs
```diff
@@ -53,9 +53,118 @@ internal static class MarshalUtils
...
+            }
+
+            var byteLength = vsnprintf(IntPtr.Zero, UIntPtr.Zero, format, args) + 1;
+            if (byteLength <= 1)
+                return string.Empty;
+
+            var buffer = IntPtr.Zero;
+            try
+            {
+                buffer = Marshal.AllocHGlobal(byteLength);
```
jeremyVignelles: If the allocation fails, you can't free IntPtr.Zero in the finally block.

Please move the allocation outside the try block.

LibVLCSharp/Shared/Helpers/MarshalUtils.cs
```diff
@@ -53,9 +53,118 @@ internal static class MarshalUtils
...
+            var byteLength = 0;
+            UseStructurePointer(listStructure, listPointer =>
+            {
+                byteLength = Native.vsnprintf_linux(IntPtr.Zero, UIntPtr.Zero, format, listPointer) + 1;
+            });
+
+            var utf8Buffer = IntPtr.Zero;
+            try
+            {
+                utf8Buffer = Marshal.AllocHGlobal(byteLength);
```
jeremyVignelles: Same here.

LibVLCSharp/Shared/Helpers/MarshalUtils.cs
```diff
@@ -53,9 +53,118 @@ internal static class MarshalUtils
...
+                Marshal.FreeHGlobal(utf8Buffer);
+            }
+        }
+
+        static string UseStructurePointer<T>(T structure, Func<IntPtr, string> action)
+        {
+            var listPointer = IntPtr.Zero;
+            try
+            {
+                listPointer = Marshal.AllocHGlobal(Marshal.SizeOf(structure));
```
jeremyVignelles: Same here.

LibVLCSharp/Shared/Helpers/MarshalUtils.cs
```diff
@@ -53,9 +53,118 @@ internal static class MarshalUtils
...
+            }
+            finally
+            {
+                Marshal.FreeHGlobal(utf8Buffer);
+            }
+        }
+
+        static string UseStructurePointer<T>(T structure, Func<IntPtr, string> action)
+        {
+            var listPointer = IntPtr.Zero;
```
jeremyVignelles: Could be renamed to `structurePointer`

LibVLCSharp/Shared/LibVLC.cs
```diff
@@ -623,28 +629,27 @@ static void UpdateProgress(IntPtr data, IntPtr dialogId, float position, string
...
-                    Task.Factory.StartNew(() => _log?.Invoke(null, new LogEventArgs(level, message, module, file, line)));
+                Task.Factory.StartNew(() => libvlc._log?.Invoke(null, new LogEventArgs(level, message, module, file, line)));
 #else
-                    Task.Run(() => _log?.Invoke(null, new LogEventArgs(level, message, module, file, line)));
+                Task.Run(() => libvlc._log?.Invoke(null, new LogEventArgs(level, message, module, file, line)));
 #endif
-                }
             }
-            finally
+            catch
```
jeremyVignelles: Why the catch-all?

LibVLCSharp/Shared/Helpers/MarshalUtils.cs
mfkl: See above ^

LibVLCSharp/Shared/Helpers/MarshalUtils.cs
mfkl: See above ^

LibVLCSharp/Shared/LibVLC.cs
mfkl: Because we don't care if it fails, it's not critical, and easy to troubleshoot as there aren't much code here. But we still want to catch exceptions should they happen.

LibVLCSharp/Shared/Helpers/MarshalUtils.cs
jeremyVignelles: I wasn't aware of that behavior of the FreeHGlobal. OK then...
However, do you think having the Alloc inside the try block helps with the OOME case?

LibVLCSharp/Shared/LibVLC.cs
jeremyVignelles: I don't know if that would be easy to detect if one out of 5000 log lines gets discarded, but OK, it shouldn't happen.

I'd be in favor of having a comment inside the catch to explain why. (in fact, that's what R# suggests when you do such things)

LibVLCSharp/Shared/Helpers/MarshalUtils.cs
mfkl: It just swallows the exception as this is a non critical failure in a specific scenario, which is the behavior we want IMO.

LibVLCSharp/Shared/LibVLC.cs
mfkl: > I'd be in favor of having a comment inside the catch to explain why

Sure

PR #67, opened 2019-09-20
Fix video lock and format callback signatures to match libvlc

lummoj:
### Description of Change ###

Fixed callback signatures of LibVLCVideoLockCb and LibVLCVideoFormatCb to match libvlc and the comment blocks above.

IntPtr planes -> out IntPtr planes as it is [OUT]
IntPtr chroma -> ref uint chroma as it is [IN/OUT] and always 4 bytes (not platform dependent)

### Issues Resolved ### 
Fixes #219 (https://code.videolan.org/videolan/LibVLCSharp/issues/219)

### API Changes ###
Changed signature of LibVLCVideoLockCb and LibVLCVideoFormatCb to match libvlc

### Platforms Affected ### 
- Core (all platforms)

None

Not applicable

### Testing Procedure ###
Tested callbacks now work by rendering an I420 video to a RV32 (RGBA) image in WPF
More testing not likely to be required as the signatures are being fixed to match libvlc

### PR Checklist ###

- [*] Rebased on top of the target branch at time of PR
- [*] Changes adhere to coding standard

### DIFF
--- a/LibVLCSharp/Shared/MediaPlayer.cs
+++ b/LibVLCSharp/Shared/MediaPlayer.cs
@@ -1741,7 +1741,7 @@ namespace LibVLCSharp.Shared
         /// <para>planes must be aligned on 32-bytes boundaries.</para>
         /// </remarks>
         [UnmanagedFunctionPointer(CallingConvention.Cdecl)]
-        public delegate IntPtr LibVLCVideoLockCb(IntPtr opaque, IntPtr planes);
+        public delegate IntPtr LibVLCVideoLockCb(IntPtr opaque, out IntPtr planes);
 
         /// <summary>Callback prototype to unlock a picture buffer.</summary>
         /// <param name="opaque">private pointer as passed to libvlc_video_set_callbacks() [IN]</param>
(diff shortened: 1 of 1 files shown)


LibVLCSharp/Shared/MediaPlayer.cs
```diff
@@ -1799,8 +1799,8 @@ public IntPtr GetFrame(out bool updated)
         /// <para>to not break assumptions that might be held by optimized code</para>
         /// <para>in the video decoders, video filters and/or video converters.</para>
         /// </remarks>
-        [UnmanagedFunctionPointer(CallingConvention.Cdecl)]
-        public delegate uint LibVLCVideoFormatCb(ref IntPtr userData, IntPtr chroma, ref uint width,
+        [UnmanagedFunctionPointer(CallingConvention.Cdecl, CharSet = CharSet.Ansi)]
```
jeremyVignelles: Why did you change the CharSet?

lummoj: Thanks for the quick reply Jeremy, apologies CharSet should be reverted. I initially tried to get this working using string and char[] but I was not be able to get it work in conjunction with ref.

I'm happy to revert the CharSet change and I can see from the sample you linked how to get the lock callback working using Marshal.WriteIntPtr so perhaps this change is unnecessary.

Can this same approach be used for the chroma in the format callback? As I don't know the native width and height I'd prefer not to use SetVideoFormat as shown in the sample.

lummoj: I found a good example for chroma here and got everything working:
https://github.com/jeremyVignelles/Vlc.DotNet/blob/develop/src/Vlc.DotNet.Core.Interops/FourCCConverter.cs
Happy to abandon this now

lummoj:
Found it was possible to achieve what I needed with existing callback signatures thanks to samples provided by @jeremyVignelles 

There is probably room for improvement with the API to make this more intuitive but I wouldn't consider it a bug.

PR #73, opened 2019-10-18
Makes the UWP VideoView more MVVM friendly

kakone:
### Description of Change ###
I replaced the SwapChainOptions and MediaPlayer properties to dependency properties. So It's easier to code in MVVM way with the UWP video view (I updated the sample to show the changes).
There is no breaking changes, property names stays the same.

Other change : the VideoView class inherits now from the Control class instead of UserControl. The developper can now updates the ControlTemplate if he wants. The default control template in Generic.xaml contains only the SwapChainPanel.

### Platforms Affected ### 
- UWP

### DIFF
--- a/LibVLCSharp/LibVLCSharp.csproj
+++ b/LibVLCSharp/LibVLCSharp.csproj
@@ -83,6 +83,9 @@ LibVLC needs to be installed separately, see VideoLAN.LibVLC.* packages.</Descri
   </ItemGroup>
   <ItemGroup Condition="$(TargetFramework.StartsWith('uap'))">
     <Compile Include="Platforms\UWP\**\*.cs" />
+    <None Update="Themes\Generic.xaml">
+      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
+    </None>
     <PackageReference Include="System.ValueTuple" Version="4.5.0" />
     <PackageReference Include="SharpDX.Direct3D11" Version="4.2.0" />
   </ItemGroup>
--- a/LibVLCSharp/LibVLCSharp.csproj
+++ b/LibVLCSharp/LibVLCSharp.csproj
@@ -56,7 +56,7 @@ LibVLC needs to be installed separately, see VideoLAN.LibVLC.* packages.</Descri
   <PropertyGroup Condition="$(NETSTANDARD_ONLY)=='true'">
     <TargetFrameworks>netstandard2.0;netstandard1.1</TargetFrameworks>
   </PropertyGroup>
-  <PropertyGroup Condition="$([MSBuild]::IsOsPlatform('Windows'))">
+  <PropertyGroup Condition="$(TargetFramework.StartsWith('uap'))">
     <GenerateLibraryLayout>true</GenerateLibraryLayout>
   </PropertyGroup>
   <ItemGroup>
--- a/LibVLCSharp/LibVLCSharp.csproj
+++ b/LibVLCSharp/LibVLCSharp.csproj
@@ -48,7 +48,6 @@ LibVLC needs to be installed separately, see VideoLAN.LibVLC.* packages.</Descri
     <PackageIcon>icon.png</PackageIcon>
     <PackageReleaseNotes>https://code.videolan.org/videolan/LibVLCSharp/blob/master/NEWS</PackageReleaseNotes>
     <PackageTags>libvlc;vlc;videolan;native;c/c++;video;audio;player;media;mediaplayer;codec;ffmpeg;xamarin;graphics;ios;android;linux;windows;macos;cross-platform</PackageTags>
-    <GenerateLibraryLayout>true</GenerateLibraryLayout>
   </PropertyGroup>
   <PropertyGroup Condition="$(UNITY_ANDROID)=='true'">
     <TargetFramework>netstandard2.0</TargetFramework>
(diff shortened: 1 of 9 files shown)


LibVLCSharp/LibVLCSharp.csproj
```diff
@@ -48,6 +48,7 @@ LibVLC needs to be installed separately, see VideoLAN.LibVLC.* packages.</Descri
     <PackageIcon>icon.png</PackageIcon>
     <PackageReleaseNotes>https://code.videolan.org/videolan/LibVLCSharp/blob/master/NEWS</PackageReleaseNotes>
     <PackageTags>libvlc;vlc;videolan;native;c/c++;video;audio;player;media;mediaplayer;codec;ffmpeg;xamarin;graphics;ios;android;linux;windows;macos;cross-platform</PackageTags>
+    <GenerateLibraryLayout>true</GenerateLibraryLayout>
```
mfkl: It used to work without this. Why is this needed now?

Samples/LibVLCSharp.UWP.Sample/MainPage.xaml
```diff
@@ -8,8 +8,11 @@
     xmlns:lvs="using:LibVLCSharp.Platforms.UWP"
     mc:Ignorable="d"
     Background="{ThemeResource ApplicationPageBackgroundThemeBrush}">
+    <Page.DataContext>
+        <local:MainViewModel />
+    </Page.DataContext>
 
     <Grid>
-      <lvs:VideoView x:Name="VideoView"></lvs:VideoView>
+      <lvs:VideoView SwapChainOptions="{Binding SwapChainOptions, Mode=TwoWay}" MediaPlayer="{Binding MediaPlayer}"></lvs:VideoView>
```
mfkl: Why `TwoWay` binding mode?

mfkl:
I get
```
Error: Cannot get 'SwapChainOptions' value (type 'Object') from type 'LibVLCSharp.UWP.Sample.MainViewModel'. BindingExpression: Path='SwapChainOptions' DataItem='LibVLCSharp.UWP.Sample.MainViewModel'; target element is 'LibVLCSharp.Platforms.UWP.VideoView' (Name='null'); target property is 'SwapChainOptions' (type 'IReferenceArray`1<String>').
```

When running the UWP sample from your branch.

kakone:
> I get
> 
> ```
> Error: Cannot get 'SwapChainOptions' value (type 'Object') from type 'LibVLCSharp.UWP.Sample.MainViewModel'. BindingExpression: Path='SwapChainOptions' DataItem='LibVLCSharp.UWP.Sample.MainViewModel'; target element is 'LibVLCSharp.Platforms.UWP.VideoView' (Name='null'); target property is 'SwapChainOptions' (type 'IReferenceArray`1<String>').
> ```
> 
> When running the UWP sample from your branch.

Not really an error because you can continue but I can easily fix it by adding a public getter in the ViewModel.

Samples/LibVLCSharp.UWP.Sample/MainPage.xaml
kakone: I would like to use `OneWayToSource` binding mode to send the `SwapChainOptions` to the ViewModel, but `OneWayToSource` binding doesn't exist in UWP. So, I used `TwoWay` binding mode.

Perhaps it might be better to create an event with the SwapChainOptions in the event data, and I can keep a non-bindable SwapChainOptions property.

LibVLCSharp/LibVLCSharp.csproj
kakone: With the add of the `Generic.xaml` file, you need this to generate the `Generic.xbf` file (XAML binary format) and place it in the good subfolder. Without this, you would get an exception. In UWP class libraries, this option is set by default.
For the other librairies (Android, iOS, ...), there is no incidence.

mfkl:
Indeed, the video didn't load for some reason first time I tried, works fine now.

> I can easily fix it by adding a public getter in the ViewModel.

Yes please :)

This swapchainoptions ugly API will go away for libvlcsharp 4.0 as we will use d3d callbacks FYI.

LibVLCSharp/LibVLCSharp.csproj
mfkl: Ok.

Samples/LibVLCSharp.UWP.Sample/MainPage.xaml
mfkl: Yes, an event will probably be better (I usually try to avoid `TwoWay` binding mode).

kakone:
Changes made, I added an Initialized event.

LibVLCSharp/LibVLCSharp.csproj
```diff
@@ -56,6 +56,9 @@ LibVLC needs to be installed separately, see VideoLAN.LibVLC.* packages.</Descri
   <PropertyGroup Condition="$(NETSTANDARD_ONLY)=='true'">
     <TargetFrameworks>netstandard2.0;netstandard1.1</TargetFrameworks>
   </PropertyGroup>
+  <PropertyGroup Condition="$([MSBuild]::IsOsPlatform('Windows'))">
+    <GenerateLibraryLayout>true</GenerateLibraryLayout>
```
mfkl: I guess this is fine, but since it's not (supposed to be) doing anything with any TFM other than the UAP one, you could have use the TFM in the condition instead of `IsOsPlatform`. As you wish though, no strong feelings here.

kakone: You are right, I changed it (cf d9d745ab2d185dfad7279e1f19a935b4750a34d6).

mfkl:
Squashed and merged https://github.com/videolan/libvlcsharp/commit/60e481bc6afc34c6c9f659561d8d43dbc89cd2bb Thank you!

PR #75, opened 2019-10-23
C#8 nullable references implementation

jeremyVignelles:
### Description of Change ###

This changes allows to use LVS with nullable annotations.
- Added a TFM for .net standard 2.1
- Updated LangVersion to 8.0
- Added `<Nullable>enable</Nullable>` in csprojs
- Implemented nullable references where applicable
- Fixed possible null references exceptions
- I also started to fix the coding style errors I was having, but once I opened a new files, a hundred more would come, so I gave up.

This is currently a Work In Progress, only LibVLCSharp and LibVLCSharp.Forms has been implemented.

### Issues Resolved ### 
- fixes https://code.videolan.org/videolan/LibVLCSharp/issues/78

### API Changes ###
Not sure if there is any. Intent will be explained more clearly.

### Platforms Affected ### 
Not applicable

When using LVS with nullable references enabled, Visual studio can warn you about a wrong usage of the API regarding null values.

Not applicable

### Testing Procedure ###
Call every API and see if it works? 😄 

### PR Checklist ###

### DIFF
--- a/LibVLCSharp.Forms.Platforms.GTK/LibVLCSharp.Forms.Platforms.GTK.csproj
+++ b/LibVLCSharp.Forms.Platforms.GTK/LibVLCSharp.Forms.Platforms.GTK.csproj
@@ -12,6 +12,7 @@ Xamarin.Forms support for other platforms are in different packages (namely LibV
     <RootNamespace>LibVLCSharp.Forms.Platforms.GTK</RootNamespace>
     <Version>3.3.1</Version>
     <LangVersion>8.0</LangVersion>
+    <Nullable>enable</Nullable>
     <PackageId>LibVLCSharp.Forms.GTK</PackageId>
     <Authors>VideoLAN</Authors>
     <Owners>VideoLAN</Owners>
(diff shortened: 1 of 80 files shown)


mfkl:
I can reproduce the macOS build failure locally with mono 6.4 and dotnet 3.0.100.

mfkl:
Whether a libvlc API can or cannot return `NULL` is not part of the libvlc API contract, e.g:
- it's not documented,
- if that changes (i.e an API that can return null becomes an API that can _never_ return null, or the opposite), it would not be considered a breaking change by the libvlc devs and versioning. Which makes our ability to keep the binding correct very hard (not just surface API changes to track anymore).

I say we should consider all strings (and possibly any pointers) possibly null. I understand it makes the C# 8 nullable support less appealing for us, but that's just the reality of native code interop.

mfkl:
What's left to be done here?

- [ ] Rebase
- [ ] use nullable for all projects

jeremyVignelles:
Just
- Rebased
- Applied what was discussed : a pointer that comes from the native is considered nullable.

I might have missed some though...

When you say "all projects", do you mean the samples as well?

mfkl:
> When you say "all projects", do you mean the samples as well?

No need to convert the sample projects, but maybe adding a small sample to libvlcsharp-samples displaying what calling a few libvlcsharp APIs with the netstandard2.1 target looks like?

mfkl:
Is this good to merge? @jeremyVignelles

kakone:
We can now remove the #nullable enable/restore lines I added in LibVLCSharp/Shared/MediaPlayerElement files waiting this PR.

jeremyVignelles:
done. I don't think there's anything left on my side. I'm ok for merging

mfkl:
🎉

PR #76, opened 2019-10-24
First PR for Uno Platform support

kakone:
### Description of Change ###

First PR to add a VideoView for Uno Platform. The next PR will add a MediaPlayerElement.

Added:
 - LibVLCSharp.Uno project
 - LibVLCSharp.Uno.Sample projects


LibVLCSharp.Uno/VideoView.iOS.cs
```diff
@@ -0,0 +1,19 @@
+﻿namespace LibVLCSharp.Platforms.UWP
```
mfkl: Don't we want a `LibVLCSharp.Uno.iOS` namespace here?

mfkl: Not that's possible...

kakone: No, the namespaces should exactly the same between the UWP version and Android or iOS. With Uno, you take your UWP code and it must work without modification in your Android and iOS app.
In fact, whe should rename the namespaces in LibVLCSharp to remove the `Platforms.UWP` but it will be a breaking change.

mfkl: Still. I don't get why you need to use the existing `LibVLCSharp.Platforms.UWP` namespace. What prevents you from putting all the Uno code in `LibVLCSharp.Uno`?

Samples/Uno/LibVLCSharp.Uno.Sample.Shared/LibVLCSharp.Uno.Sample.Shared.shproj
```diff
@@ -0,0 +1,13 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
```
mfkl: Hmm the Uno people seem to like their shared projects :D

LibVLCSharp.Uno/VideoView.iOS.cs
kakone: Because there is no "Uno" version of the VideoView for UWP, we reuse the LibVLCSharp.Platforms.UWP.VideoView.

kakone: No, you can't do this. Otherwise, the LibVLCSharp.Uno.Sample.UWP project will not compile.

Samples/Uno/LibVLCSharp.Uno.Sample.Shared/LibVLCSharp.Uno.Sample.Shared.shproj
kakone: Yes, I don't like too but it's the default template when you create an Uno project, I preferred to let it like that.

mfkl: Sure, just making an observation :+1:

LibVLCSharp.Uno/LibVLCSharp.Uno.csproj
```diff
@@ -0,0 +1,44 @@
...
+    <Version>0.1.1-alpha</Version>
+    <GeneratePackageOnBuild>false</GeneratePackageOnBuild>
+    <Description>VLC MediaPlayerElement for Uno Platform (UWP, Android, iOS)</Description>
+    <PackageTags>libvlc;vlc;videolan;video;audio;player;media;mediaplayer;mediaelement;mediaplayerelement;codec;ffmpeg;graphics;ios;android;windows;cross-platform;uno</PackageTags>
+    <PackageProjectUrl>https://github.com/libvlcsharp/libvlcsharp.uno</PackageProjectUrl>
+    <PackageId>LibVLCSharp.Uno</PackageId>
+    <PackageLicenseExpression>LGPL-2.1-or-later</PackageLicenseExpression>
+  </PropertyGroup> 
+  <ItemGroup Condition="'$(TargetFramework)'=='xamarinios10' or '$(TargetFramework)'=='monoandroid81' or '$(TargetFramework)'=='monoandroid90'">
+    <PackageReference Include="Uno.UI" Version="2.0.512-dev.3565" />
```
mfkl: Do you need that pre-release version specifically?

kakone: No, I'm taking the last pre-release of version 2, waiting a stable version.

LibVLCSharp.Uno/VideoView.iOS.cs
mfkl: Indeed, but it should still be possible. Also why not provide a UAP build? reactiveui seems to have one (along with a custom namespace) https://github.com/reactiveui/ReactiveUI/blob/master/src/ReactiveUI.Uno/ReactiveUI.Uno.csproj

kakone: Because in our case, I don't need anything from `LibVLCSharp.Uno` when I create an UWP application.
An UWP developer will not include `LibVLCSharp.Uno` because he doesn't need it, `LibVLCSharp` is enough.
It's only when he wants to port its application on Android or iOS that he will use `LibVLCSharp.Uno`. No work to do, he takes the UWP code and it must work without modifying the namespaces.

LibVLCSharp.Uno/VideoView.cs
```diff
@@ -0,0 +1,110 @@
...
+        public string[] SwapChainOptions
+        {
+            get
+            {
+                if (!IsLoaded)
+                {
+                    throw new InvalidOperationException("You must wait for the VideoView to be loaded before calling GetSwapChainOptions()");
+                }
+
+                return Array.Empty<string>();
```
mfkl: This property either throws or returns an empty array. Can we do without?

LibVLCSharp.Uno/VideoView.iOS.cs
mfkl: Still... :)
Re-using `LibVLCSharp.UWP` directly wrecks the namespace for LibVLCSharp with Uno on Android and iOS. This will be very confusing for users and I don't see the benefit.
We should provide a UAP build, even if it is only a thin shell over `LibVLCSharp.UWP`, and that will allow proper namespace usage.

LibVLCSharp.Uno/VideoView.Android.cs
```diff
@@ -0,0 +1,21 @@
+﻿using Uno.UI;
+
+namespace LibVLCSharp.Platforms.UWP
+{
+    /// <summary>
+    /// Video view
+    /// </summary>
+    public partial class VideoView
```
mfkl: I try to avoid `partial class` as much as possible. Do you think we can come up with something clean without it?

LibVLCSharp.Uno/LibVLCSharp.Uno.csproj
```diff
@@ -0,0 +1,44 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<Project Sdk="MSBuild.Sdk.Extras/2.0.54">
+  <PropertyGroup>
+    <LangVersion>latest</LangVersion>
```
mfkl: This should be pinned to 8.0.

Samples/Uno/LibVLCSharp.Uno.Sample.UWP/LibVLCSharp.Uno.Sample.UWP.csproj
```diff
@@ -0,0 +1,146 @@
...
+			This is not an issue when libraries are referenced through nuget packages. See https://github.com/nventive/Uno/issues/446 for more details.
+			-->
+      <Version>6.2.9</Version>
+    </PackageReference>
+    <PackageReference Include="Microsoft.Extensions.Logging.Console" Version="1.1.2" />
+    <PackageReference Include="Microsoft.Extensions.Logging.Filter" Version="1.1.2" />
+    <PackageReference Include="Microsoft.Xaml.Behaviors.Uwp.Managed">
+      <Version>2.0.1</Version>
+    </PackageReference>
+    <PackageReference Include="MvvmLightLibs">
```
mfkl: I like having the most minimal sample projects in this repo as it makes for a good setup template. Even if easily changed, defaulting to a specific mvvm framework makes the startup sample less generic / easy to start with.

kakone: Done in 5af453ca36106c610903712c37ceed2bf8e15710

LibVLCSharp.Uno/LibVLCSharp.Uno.csproj
kakone: Done in f30631a566c07c0a98001abcb25665e305a869a5

LibVLCSharp.Uno/VideoView.Android.cs
kakone: I updated the code in d481e7f32452bb8653f34b6723aaa4e523c2fca4. I use an abstract generic base class instead of a partial class.
But even if my class is no longer partial, I still have to let the `partial` keyword for Uno code generation.

LibVLCSharp.Uno/VideoView.cs
kakone: No, I can remove the `throw` but I must let this property. The goal is to be able to use exactly the same code with UWP, Android or iOS.

LibVLCSharp.Uno/VideoView.iOS.cs
kakone: Still too... :)
Yes, I can do this but I don't think it's a good idea. Even there is no benefit, it's not the "Uno Platform spirit" :)
In fact, Uno doesn't exist for UWP and you should not have any Uno library in your UWP project. You have your existing current code for UWP, and it should remain the same when you want to port your app in Android or iOS.
IMO, the real change that should be done is to remove this .Platform.XXX in the namespace in the LibVLCSharp library because it's useless. But, this would introduce a breaking change.

LibVLCSharp.Uno/VideoView.Android.cs
kakone: For more explanation on the Uno code generation, you can look at [this post](https://medium.com/@unoplatform/talkin-bout-my-generation-how-the-uno-platform-generates-code-part-1-under-the-hood-7664d83c4f90).

LibVLCSharp.Uno/VideoView.iOS.cs
kakone: Finally, I was thinking about where I will place the MediaPlayerElement. As the code will be exactly the same for all platforms, I will not doing one library for UWP and another for Uno Android and iOS. The best place for the code of the MediaPlayerElement will be the LibVLCSharp.Uno project. So, in this case, we will be obliged to add a reference of the LibVLCSharp.Uno in our UWP project. I'm going to make an exception by adding an UWP target in this Uno library and take the opportunity to change the namespace (for UWP, I will just inherit of the VideoView and change the namespace).

kakone: I did it in e218b978878e1635391cb8297d2375210ba0ee6a.

mfkl:
Enabling
```
<GenerateDocumentationFile>true</GenerateDocumentationFile>
<TreatWarningsAsErrors>true</TreatWarningsAsErrors>
```
Fails the build with
```
obj\Debug\monoandroid90\90\g\NativeCtorsGenerator\LibVLCSharp_Uno_VideoView_8619a54875bbd4210ec7ce9c8389e9dc.g.cs(16,10,16,19): error CS1591: Missing XML comment for publicly visible type or member 'VideoView.VideoView()'
1>obj\Debug\monoandroid90\90\g\XamlCodeGenerator\GlobalStaticResources.g.cs(38,22,38,32): error CS1591: Missing XML comment for publicly visible type or member 'GlobalStaticResources.Initialize()'
1>obj\Debug\monoandroid90\90\g\XamlCodeGenerator\VideoView_156923378629bd4eb274fde7d844c5f8.g.cs(51,47,51,88): error CS1591: Missing XML comment for publicly visible type or member 'GlobalStaticResources.__ImplicitStyle_LibVLCSharp_Uno_VideoView'
```
Likely a Uno issue. Let's disregard that for now.

LibVLCSharp.Uno/LibVLCSharp.Uno.csproj
```diff
@@ -0,0 +1,51 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<Project Sdk="MSBuild.Sdk.Extras/2.0.54">
+  <PropertyGroup>
+    <LangVersion>8.0</LangVersion>
+    <Nullable>enable</Nullable>
+    <TargetFrameworks>xamarinios10;monoandroid81;monoandroid90;uap10.0.16299</TargetFrameworks>
```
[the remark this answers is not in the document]
kakone: When you create an Uno project, there are the two targets. I don't know really why but I think the generated code is not the same in some cases. Perhaps it's not necessary in our case, I'm going to look at that.

Samples/Uno/LibVLCSharp.Uno.Sample.Shared/App.xaml.cs
```diff
@@ -0,0 +1,145 @@
...
+            //TODO: Save application state and stop any background activity
+            deferral.Complete();
+        }
+
+
+        /// <summary>
+        /// Configures global logging
+        /// </summary>
+        /// <param name="factory"></param>
+        static void ConfigureFilters(ILoggerFactory factory)
```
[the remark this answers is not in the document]
kakone: Yes, I can remove it (this code is in the default Uno template).

kakone:
> Enabling
> 
> ```
> <GenerateDocumentationFile>true</GenerateDocumentationFile>
> <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
> ```
> 
> Fails the build with
> 
> ```
> obj\Debug\monoandroid90\90\g\NativeCtorsGenerator\LibVLCSharp_Uno_VideoView_8619a54875bbd4210ec7ce9c8389e9dc.g.cs(16,10,16,19): error CS1591: Missing XML comment for publicly visible type or member 'VideoView.VideoView()'
> 1>obj\Debug\monoandroid90\90\g\XamlCodeGenerator\GlobalStaticResources.g.cs(38,22,38,32): error CS1591: Missing XML comment for publicly visible type or member 'GlobalStaticResources.Initialize()'
> 1>obj\Debug\monoandroid90\90\g\XamlCodeGenerator\VideoView_156923378629bd4eb274fde7d844c5f8.g.cs(51,47,51,88): error CS1591: Missing XML comment for publicly visible type or member 'GlobalStaticResources.__ImplicitStyle_LibVLCSharp_Uno_VideoView'
> ```
> 
> Likely a Uno issue. Let's disregard that for now.

OK, we can disregard that for now but I will create an issue (and perhaps a PR) in the Uno project to resolve this.

kakone:
> Running the UWP (not Uno) sample gives
> 
> ```
> Exception thrown: 'Windows.UI.Xaml.Markup.XamlParseException' in LibVLCSharp.UWP.Sample.exe
> WinRT information: Failed to assign to property 'LibVLCSharp.Platforms.UWP.VideoViewBase.MediaPlayer'. [Line: 18 Position: 22]
> An exception of type 'Windows.UI.Xaml.Markup.XamlParseException' occurred in LibVLCSharp.UWP.Sample.exe but was not handled in user code
> WinRT information: Failed to assign to property 'LibVLCSharp.Platforms.UWP.VideoViewBase.MediaPlayer'. [Line: 18 Position: 22]
> The text associated with this error code could not be found.
> 
> Failed to assign to property 'LibVLCSharp.Platforms.UWP.VideoViewBase.MediaPlayer'. [Line: 18 Position: 22]
> ```
> 
> I'm afraid the new base class is kind of a breaking change for UWP, but we will have to do with it.

I'm surprised that this bug doesn't exist in Uno.UWP sample. It's an error on the type when I register the MediaPlayer dependency property. I'm going to correct this.

Samples/Uno/LibVLCSharp.Uno.Sample.Shared/App.xaml.cs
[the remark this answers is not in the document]
kakone: Removed by ff60d538a7d7c849c5918c7f3309b044b81b1ff4

LibVLCSharp.Uno/LibVLCSharp.Uno.csproj
```diff
@@ -0,0 +1,51 @@
...
+    <!-- Ensures the .xr.xml files are generated in a proper layout folder -->
+    <GenerateLibraryLayout>true</GenerateLibraryLayout>
+    <Authors>VideoLAN</Authors>
+    <Owners>VideoLAN</Owners>
+    <Version>0.1.1-alpha</Version>
+    <GeneratePackageOnBuild>false</GeneratePackageOnBuild>
+    <Description>VLC MediaPlayerElement for Uno Platform (UWP, Android, iOS)</Description>
+    <PackageTags>libvlc;vlc;videolan;video;audio;player;media;mediaplayer;mediaelement;mediaplayerelement;codec;ffmpeg;graphics;ios;android;windows;cross-platform;uno</PackageTags>
+    <PackageProjectUrl>https://github.com/libvlcsharp/libvlcsharp.uno</PackageProjectUrl>
```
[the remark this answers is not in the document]
kakone: OK, you can do it later :)

LibVLCSharp.Uno/LibVLCSharp.Uno.csproj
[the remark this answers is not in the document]
kakone: I removed it in edf54504a8356122a0e7099c28a2acd345b50d64 because I didn't see any reason to add it.

kakone:
It's good for me. I will add `<GenerateDocumentationFile>true</GenerateDocumentationFile>` later when the [PR](https://github.com/unoplatform/uno/issues/1983) will be merged.

mfkl:
Merged, thanks for the contribution 🎈

PR #77, opened 2019-10-31
Adds MediaPlayerElement for Uno Platform

kakone:
### Description of Change ###

Adds MediaPlayerElement and MediaTransportControls for Uno Platform

### PR Checklist ###
For the first version :
- [x] Show/Hide playback controls automatically
- [x] Play/Pause button
- [x] Seekbar
- [x] Volume / Mute
- [x] Closed captions selection
- [x] Audio tracks selection
- [x] Stop button
- [x] Zoom button
- [x] Cast button
- [x] Errors display
- [x] Lifecycle management
- [x] Do an unique MediaTransportControls.xaml template
- [x] Buttons content dependency properties
- [x] Compact mode
- [x] Keep screen on

Code factoring with LibVLCSharp.Forms.MediaPlayerElement / PlaybackControls :
- [x] AspectRatioManager
- [x] AutoHideNotifier
- [x] DeviceAwakeningManager
- [x] VolumeManager
- [x] SeekbarManager
- [x] AudioTracksManager
- [x] SubtitlesManager
- [x] CastManager
- [x] BufferingProgressNotifier
- [x] StateManager


kakone:
@mfkl, where do you think I should put the common code between Uno MediaPlayerElement and Xamarin.Forms MediaPlayerElement ?
Directly in the libvlcsharp project ? In a libvlcsharp.Shared.MediaPlayerElement namespace ?
Or in another libvlcsharp.MediaPlayerElement project ?

mfkl:
Good question.

Putting the shared `MediaElement` code in the `LibVLCSharp.Forms` package would mean `LibVLCSharp.Uno` would have an indirect dependency on Xamarin.Forms. I don't think we want that.

If we don't create an additional package, the `LibVLCSharp` package seems to be the only option.

Although when making this design decision, we have to keep in mind that per-platform MediaElement implementations might come in later. (https://code.videolan.org/videolan/LibVLCSharp/issues/183 and https://code.videolan.org/videolan/LibVLCSharp/issues/182, notably)

kakone:
I'm separating each feature of the MediaPlayerElement in several managers. Each manager can also be used independently, without using the MediaPlayerElement or the PlaybackControls.
So, putting them in the LibVLCSharp package seems to be a good choice.

mfkl:
> I'm separating each feature of the MediaPlayerElement in several managers. Each manager can also be used independently, without using the MediaPlayerElement or the PlaybackControls.

Hmm I didn't expect that this PR would take this turn :-) 

I don't think it's a good idea, because the initial goal of libvlcsharp is to stay as close as the libvlc API (like libvlcpp/libvlcppcx). The manager abstraction layer, much like Vlc.DotNet, brings little convenience IMO to the end user while removing the libvlc APIs proximity. It also adds more code, which has to be maintained, fix bugs, and potentially manage an entire new layer of object with native interop principles in mind. 

We also try to [avoid](https://blog.codinghorror.com/i-shall-call-it-somethingmanager/) using `Manager` suffixes as much as possible (aside from events managers I guess).

I do agree we need a way to do composition for the media element shared code so we can re-use. But those APIs cannot be public.

kakone:
The first goal of the "managers" (I don't know how I can name them) is to facilitate the maintenance and reuse of the code in the different MediaPlayerElements (the code is really much more understandable like that).
I agree to not expose them publicly (for not having another API to publicly maintain) but I don't know how I can do that. These classes need to be public to be reutilizable in the different assemblies. If I put them in a "Internals" namespace and clearly indicates that this code is for internal use, will that be enough? Do you have another idea?

kakone:
I can put an [InternalsVisibleTo] attribute, I don't know if it's a good solution.

mfkl:
> I can put an [InternalsVisibleTo] attribute, I don't know if it's a good solution.

Hmm I'm not sure, we'd then probably need to provide 2 versions of the lib (for signed and unsigned.

> If I put them in a "Internals" namespace and clearly indicates that this code is for internal use, will that be enough?

No, users will happily ignore that and use it anyway.

>  Do you have another idea?

I'm thinking. Is it possible to do some smart composition with the `protected internal` access modifier?

mfkl:
`InternalsVisibleTo` might be a good idea.

jeremyVignelles:
I'm more in favor of an internal namespace. I fear that InternalsVisibleTo creates a dependency complexity.

If the users want to use that anyway, that's their problem, but they are aware that anything might break there without notice.

LibVLCSharp.Uno/SuspensionManager.cs
```diff
@@ -0,0 +1,52 @@
+﻿using LibVLCSharp.Shared;
+using Windows.Storage;
+
+namespace LibVLCSharp.Uno
+{
+    /// <summary>
+    /// Suspension manager
+    /// </summary>
+    public static class SuspensionManager
```
jeremyVignelles: I tend to call my classes that contains only static functions "helpers", so `SuspensionHelpers` ?

LibVLCSharp/Shared/MediaPlayerElement/AutoHideManager.cs
```diff
@@ -0,0 +1,157 @@
+﻿#nullable enable
+using System;
+using System.Threading;
+
+namespace LibVLCSharp.Shared.MediaPlayerElement
+{
+    /// <summary>
+    /// Manage the playback controls auto hide feature
+    /// </summary>
+    public class AutoHideManager : MediaPlayerElementManagerBase
```
jeremyVignelles: `AutoHideNotifier`?

kakone:
> I'm more in favor of an internal namespace. I fear that InternalsVisibleTo creates a dependency complexity.
> 
> If the users want to use that anyway, that's their problem, but they are aware that anything might break there without notice.

I don't disagree. For now, I put an attribute [InternalsVisibleTo] but I can go back. I let @mfkl make the decision.

kakone:
I'm think I'm done (if you have ideas to name things a little better instead of 'managers', they are welcome).
I noticed some bugs but it does not come from this code : sound doesn't work in UWP with LibVLC.UWP 3.2.0 (cf [LibVLCSharp/issues/253](https://code.videolan.org/videolan/LibVLCSharp/issues/253)). In 3.1.1, the Mute property always returns false, even if the sound is muted.
With Android, there are some crashes when I stop the MediaPlayer or when the end of the media is reached (perhaps due to this bug : https://github.com/xamarin/xamarin-android/issues/2893).
And I can't test with iOS (I don't have a Mac, neither iPhone or iPad).

mfkl:
This is a lot of code to review, it will take me a while so please bear with me :-)

Quickly tried to run iOS, the UI shows up but not the video. Not much logs besides:

```
2019-11-18 06:24:56.664708+0100 Sample.MediaPlayerElement[62738:844067] SecTaskLoadEntitlements failed error=22 cs_flags=200, pid=62738
2019-11-18 06:24:56.664950+0100 Sample.MediaPlayerElement[62738:844067] SecTaskCopyDebugDescription: Sample.MediaPlay[62738]/0#-1 LF=0
2019-11-18 06:24:56.673021+0100 Sample.MediaPlayerElement[62738:844067] SecTaskLoadEntitlements failed error=22 cs_flags=200, pid=62738
2019-11-18 06:24:56.673250+0100 Sample.MediaPlayerElement[62738:844067] SecTaskCopyDebugDescription: Sample.MediaPlay[62738]/0#-1 LF=0
[0:] The Bindable attribute is missing and the type [Windows.UI.Xaml.UnsetValue] is not known by the MetadataProvider. Reflection was used instead of the binding engine and generated static metadata. Add the Bindable attribute to prevent this message and performance issues.
[0:] The Bindable attribute is missing and the type [Windows.UI.Xaml.UnsetValue] is not known by the MetadataProvider. Reflection was used instead of the binding engine and generated static metadata. Add the Bindable attribute to prevent this message and performance issues.
[0:] The Bindable attribute is missing and the type [Windows.UI.Xaml.UnsetValue] is not known by the MetadataProvider. Reflection was used instead of the binding engine and generated static metadata. Add the Bindable attribute to prevent this message and performance issues.
[0:] The Bindable attribute is missing and the type [Windows.UI.Xaml.UnsetValue] is not known by the MetadataProvider. Reflection was used instead of the binding engine and generated static metadata. Add the Bindable attribute to prevent this message and performance issues.
[0:] The Bindable attribute is missing and the type [Windows.UI.Xaml.UnsetValue] is not known by the MetadataProvider. Reflection was used instead of the binding engine and generated static metadata. Add the Bindable attribute to prevent this message and performance issues.
[0:] The Bindable attribute is missing and the type [Windows.UI.Xaml.UnsetValue] is not known by the MetadataProvider. Reflection was used instead of the binding engine and generated static metadata. Add the Bindable attribute to prevent this message and performance issues.
[0:] The Bindable attribute is missing and the type [Sample.MediaPlayerElement.MainViewModel] is not known by the MetadataProvider. Reflection was used instead of the binding engine and generated static metadata. Add the Bindable attribute to prevent this message and performance issues.
[0:] The Bindable attribute is missing and the type [Sample.MediaPlayerElement.MainViewModel] is not known by the MetadataProvider. Reflection was used instead of the binding engine and generated static metadata. Add the Bindable attribute to prevent this message and performance issues.
[0:] The Bindable attribute is missing and the type [Sample.MediaPlayerElement.MainViewModel] is not known by the MetadataProvider. Reflection was used instead of the binding engine and generated static metadata. Add the Bindable attribute to prevent this message and performance issues.
```

I will try to have a look later (unless you know the fix already).

mfkl:
> I don't disagree. For now, I put an attribute [InternalsVisibleTo] but I can go back. I let @mfkl make the decision.

I think `InternalsVisibleTo` is the way to go.

kakone:
> This is a lot of code to review, it will take me a while so please bear with me :-)
> 
> Quickly tried to run iOS, the UI shows up but not the video. Not much logs besides:
> 
> ```
> 2019-11-18 06:24:56.664708+0100 Sample.MediaPlayerElement[62738:844067] SecTaskLoadEntitlements failed error=22 cs_flags=200, pid=62738
> 2019-11-18 06:24:56.664950+0100 Sample.MediaPlayerElement[62738:844067] SecTaskCopyDebugDescription: Sample.MediaPlay[62738]/0#-1 LF=0
> 2019-11-18 06:24:56.673021+0100 Sample.MediaPlayerElement[62738:844067] SecTaskLoadEntitlements failed error=22 cs_flags=200, pid=62738
> 2019-11-18 06:24:56.673250+0100 Sample.MediaPlayerElement[62738:844067] SecTaskCopyDebugDescription: Sample.MediaPlay[62738]/0#-1 LF=0
> [0:] The Bindable attribute is missing and the type [Windows.UI.Xaml.UnsetValue] is not known by the MetadataProvider. Reflection was used instead of the binding engine and generated static metadata. Add the Bindable attribute to prevent this message and performance issues.
> [0:] The Bindable attribute is missing and the type [Windows.UI.Xaml.UnsetValue] is not known by the MetadataProvider. Reflection was used instead of the binding engine and generated static metadata. Add the Bindable attribute to prevent this message and performance issues.
> [0:] The Bindable attribute is missing and the type [Windows.UI.Xaml.UnsetValue] is not known by the MetadataProvider. Reflection was used instead of the binding engine and generated static metadata. Add the Bindable attribute to prevent this message and performance issues.
> [0:] The Bindable attribute is missing and the type [Windows.UI.Xaml.UnsetValue] is not known by the MetadataProvider. Reflection was used instead of the binding engine and generated static metadata. Add the Bindable attribute to prevent this message and performance issues.
> [0:] The Bindable attribute is missing and the type [Windows.UI.Xaml.UnsetValue] is not known by the MetadataProvider. Reflection was used instead of the binding engine and generated static metadata. Add the Bindable attribute to prevent this message and performance issues.
> [0:] The Bindable attribute is missing and the type [Windows.UI.Xaml.UnsetValue] is not known by the MetadataProvider. Reflection was used instead of the binding engine and generated static metadata. Add the Bindable attribute to prevent this message and performance issues.
> [0:] The Bindable attribute is missing and the type [Sample.MediaPlayerElement.MainViewModel] is not known by the MetadataProvider. Reflection was used instead of the binding engine and generated static metadata. Add the Bindable attribute to prevent this message and performance issues.
> [0:] The Bindable attribute is missing and the type [Sample.MediaPlayerElement.MainViewModel] is not known by the MetadataProvider. Reflection was used instead of the binding engine and generated static metadata. Add the Bindable attribute to prevent this message and performance issues.
> [0:] The Bindable attribute is missing and the type [Sample.MediaPlayerElement.MainViewModel] is not known by the MetadataProvider. Reflection was used instead of the binding engine and generated static metadata. Add the Bindable attribute to prevent this message and performance issues.
> ```
> 
> I will try to have a look later (unless you know the fix already).

I did a little change on VideoView.iOS.cs file (by taking inspiration from this [source code]( https://github.com/unoplatform/uno/blob/master/src/Uno.UI/UI/Xaml/Controls/MediaPlayer/MediaPlayerPresenter.iOS.cs)). Maybe that will change something.
The strange warnings 'Bindable attribute is missing' are also present on Android, I am trying to understand their meaning.

mfkl:
Still an issue. I'll troubleshoot when I have time (probably next week).

LibVLCSharp.Uno/LibVLCSharp.Uno.csproj
```diff
@@ -46,19 +49,32 @@ It also contains a VLC MediaPlayerElement for the Uno Platform (UWP, Android, iO
...
     <Compile Include="**\*.Android.*cs" />
+    <Page Include="**\*.Android.*xaml" Exclude="bin\**\*.xaml;obj\**\*.xaml" />
   </ItemGroup>
   <ItemGroup Condition="$(TargetFramework.StartsWith('xamarinios'))">
     <Compile Include="**\*.iOS.*cs" />
+    <Page Include="**\*.iOS.*xaml" Exclude="bin\**\*.xaml;obj\**\*.xaml" />
+  </ItemGroup>
+  <ItemGroup>
+    <Reference Include="System.ComponentModel.DataAnnotations">
+      <HintPath>C:\Program Files (x86)\Reference Assemblies\Microsoft\Framework\Xamarin.iOS\v1.0\System.ComponentModel.DataAnnotations.dll</HintPath>
```
mfkl: This will be a problem for building on macOS I'm afraid.

mfkl:
The reason it doesn't work on iOS is because the `Microsoft.Xaml.Interactivity` and `Microsoft.Xaml.Interactions.Core` packages used in the MediaElement sample seem to not work on iOS; if I call Initialize myself, it works.

I'd suggest to remove them from the samples:
- https://www.nuget.org/packages/Uno.Microsoft.Xaml.Behaviors.Uwp.Managed/ for example isn't really maintained it seems (2 pre-release nugets only).
- Let's keep the samples the simplest.

I think you should also rename the samples to `LibVLCSharp.Uno.MediaPlayerElement.Samples.iOS|Android` for aligning with the rest of the samples names.

Samples/Uno/Sample.MediaPlayerElement/Sample.MediaPlayerElement.iOS/Sample.MediaPlayerElement.iOS.csproj
```diff
@@ -0,0 +1,168 @@
...
+    <Reference Include="System.Core" />
+    <Reference Include="Xamarin.iOS" />
+  </ItemGroup>
+  <ItemGroup>
+    <PackageReference Include="Uno.Microsoft.Xaml.Behaviors.Uwp.Managed">
+      <Version>2.0.1-uno.28</Version>
+    </PackageReference>
+    <PackageReference Include="Uno.UI" Version="2.0.512-dev.3644" />
+    <PackageReference Include="Microsoft.Extensions.Logging.Console" Version="1.1.2" />
+    <PackageReference Include="Microsoft.Extensions.Logging.Filter" Version="1.1.2" />
```
mfkl: I think we don't need those references.

LibVLCSharp.Uno/LibVLCSharp.Uno.csproj
kakone: Visual Studio does some weird stuff sometimes. Thank you, I delete it.

Samples/Uno/Sample.MediaPlayerElement/Sample.MediaPlayerElement.iOS/Sample.MediaPlayerElement.iOS.csproj
kakone: OK, I remove them.

kakone:
> I'd suggest to remove them from the samples:
> 
> * https://www.nuget.org/packages/Uno.Microsoft.Xaml.Behaviors.Uwp.Managed/ for example isn't really maintained it seems (2 pre-release nugets only).
> * Let's keep the samples the simplest.

OK, I removed the packages.

> I think you should also rename the samples to `LibVLCSharp.Uno.MediaPlayerElement.Samples.iOS|Android` for aligning with the rest of the samples names.

I can't. I shortcut the name because I had a problem with Uno code generation ("Path too long").

mfkl:
Is this good to merge?

kakone:
It's good for me

mfkl:
🎉

PR #83, opened 2019-12-14
VolumeChanged event bug workaround

kakone:
### Description of Change ###
Currently, on Android, if I subscribe to the MediaPlayer.VolumeChanged event, the application crashes when the media player is stopped (the problem must come from libVLC or Xamarin).
This pull request removes the subscription to the MediaPlayer.VolumeChanged event. It's just a workaround, idealy, it shouldn't crash when the media player is stopped.

### DIFF
--- a/LibVLCSharp.Uno/PlaybackControlsBase.cs
+++ b/LibVLCSharp.Uno/PlaybackControlsBase.cs
@@ -74,7 +74,7 @@ namespace LibVLCSharp.Uno
             castRenderersDiscoverer.Enabled = IsCastButtonVisible;
             var volumeManager = Manager.Get<VolumeManager>();
             volumeManager.EnabledChanged += (sender, e) => VolumeMuteButtonAvailabilityCommand?.Update();
-            volumeManager.VolumeChanged += (sender, e) => UpdateVolumeSlider();
+            //volumeManager.VolumeChanged += (sender, e) => UpdateVolumeSlider();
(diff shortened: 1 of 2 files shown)


mfkl:
Argh... this is getting really annoying. Need to find a fix.

Can you share the stacktrace? I'm assuming it's the same than the Vout event crash...

kakone:
Here is the output when it crashes :
```
D/Mono    ( 8923): DllImport searching in: 'libvlc' ('libvlc.so').
D/Mono    ( 8923): Searching for 'libvlc_media_player_stop'.
D/Mono    ( 8923): Probing 'libvlc_media_player_stop'.
D/Mono    ( 8923): Found as 'libvlc_media_player_stop'.
E/NdkMediaCodec( 8923): sf error code: -38
D/SurfaceUtils( 8923): disconnecting from surface 0x7546059010, reason disconnectFromSurface
W/AMessage( 8923): failed to deliver message as target handler 2 is gone.
F/lumechangedtes( 8923): java_vm_ext.cc:542] JNI DETECTED ERROR IN APPLICATION: a thread (tid 8959 is making JNI calls without being attached
F/lumechangedtes( 8923): java_vm_ext.cc:542]     in call to GetJavaVM
```
I did a simple repro project in Xamarin.Android here : https://github.com/kakone/VolumeChangedTest

mfkl:
I guess I'll have to take this change in until I fix the bug in xamarin-android. It's interesting that it repro consistently, and only for 2 event types and not the others.

PR #84, opened 2019-12-23
Fix collected callbacks

jeremyVignelles:
### Description of Change ###
We have seen several issues related to libvlc callbacks that were called while the .net GC has collected the methods.

This fixes those kind of issues in the following APIs:
- Video callbacks
- Audio callbacks
- Dialog callbacks

### Issues Resolved ### 
- fixes #275

### API Changes ###
None, all changes are internal

### Platforms Affected ### 
- Core (all platforms)

None

Not applicable

### Testing Procedure ###
https://github.com/sergiocapozzi77/SetDialogsCrash
https://github.com/jeremyVignelles/libvlcsharp-samples/commit/1ed62bd769ea16860f161dbc12911dae88c278f0

### PR Checklist ###

### DIFF
--- a/LibVLCSharp/Shared/LibVLC.cs
+++ b/LibVLCSharp/Shared/LibVLC.cs
@@ -531,6 +531,14 @@ namespace LibVLCSharp.Shared
 
         #region DialogManagement
 
+        private static readonly DisplayErrorCallback InternalErrorCallback = Error;
+        private static readonly DisplayLoginCallback InternalLoginCallback = Login;
+        private static readonly DisplayQuestionCallback InternalQuestionCallback = Question;
+        private static readonly DisplayProgressCallback InternalDisplayProgressCallback = DisplayProgress;
+        private static readonly CancelCallback InternalCancelCallback = Cancel;
+        private static readonly UpdateProgressCallback InternalUpdateProgressCallback = UpdateProgress;
+        private static readonly DialogCallbacks DialogCbs = new DialogCallbacks(InternalErrorCallback, InternalLoginCallback, InternalQuestionCallback, InternalDisplayProgressCallback, InternalCancelCallback, InternalUpdateProgressCallback);
+
         /// <summary>
         /// Register callbacks in order to handle VLC dialogs. 
         /// LibVLC 3.0.0 and later.
@@ -551,8 +559,7 @@ namespace LibVLCSharp.Shared
             _displayProgress = displayProgress ?? throw new ArgumentNullException(nameof(displayProgress));
             _updateProgress = updateProgress ?? throw new ArgumentNullException(nameof(updateProgress));
 
-            _dialogCbs = new DialogCallbacks(Error, Login, Question, DisplayProgress, Cancel, UpdateProgress);
-            Native.LibVLCDialogSetCallbacks(NativeReference, _dialogCbs, IntPtr.Zero);
+            Native.LibVLCDialogSetCallbacks(NativeReference, DialogCbs, IntPtr.Zero);
         }
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


LibVLCSharp/Shared/LibVLC.cs
```diff
@@ -531,6 +531,14 @@ void UnsetLog()
 
         #region DialogManagement
 
+        private static readonly DisplayErrorCallback InternalErrorCallback = Error;
+        private static readonly DisplayLoginCallback InternalLoginCallback = Login;
+        private static readonly DisplayQuestionCallback InternalQuestionCallback = Question;
+        private static readonly DisplayProgressCallback InternalDisplayProgressCallback = DisplayProgress;
+        private static readonly CancelCallback InternalCancelCallback = Cancel;
+        private static readonly UpdateProgressCallback InternalUpdateProgressCallback = UpdateProgress;
+        private static readonly DialogCallbacks DialogCbs = new DialogCallbacks(InternalErrorCallback, InternalLoginCallback, InternalQuestionCallback, InternalDisplayProgressCallback, InternalCancelCallback, InternalUpdateProgressCallback);
```
mfkl: Why not new the struct `SetDialogHandlers`?

LibVLCSharp/Shared/LibVLC.cs
```diff
@@ -562,8 +569,7 @@ public void UnsetDialogHandlers()
         {
             if (DialogHandlersSet)
             {
-                _dialogCbs = default;
-                Native.LibVLCDialogSetCallbacks(NativeReference, _dialogCbs, IntPtr.Zero);
+                Native.LibVLCDialogSetCallbacks(NativeReference, default, IntPtr.Zero);
```
mfkl: Should we set the struct to `default` as well?

LibVLCSharp/Shared/LibVLC.cs
jeremyVignelles: The struct will never change since it uses static callbacks. That's why.
Maybe we don't need to keep the struct once it's passed to the method? But I didn't want to change the behavior too much.

LibVLCSharp/Shared/LibVLC.cs
jeremyVignelles: passing `default` works as is. The structure doesn't need to change.

jeremyVignelles:
Rebased and reverted nullables.

Dialogs callbacks tested with that code in the netcore sample:

```cs
using System;
using System.Threading.Tasks;
using LibVLCSharp.Shared;

namespace LibVLCSharp.NetCore.Sample
{
    class Program
    {
        static void Main(string[] args)
        {
            Core.Initialize();

            using(var libVLC = new LibVLC())
            {
                SetDialogHandlers(libVLC);
                GC.Collect();
                var media = new Media(libVLC, "http://commondatastorage.googleapis.com/INEXISTANT", FromType.FromLocation);
                using (var mp = new MediaPlayer(media))
                {
                    media.Dispose();
                    mp.Play();
                    Console.ReadKey();
                }
                UnsetDialogHandlers(libVLC);
            }
        }

        private static void SetDialogHandlers(LibVLC libVLC)
        {
            libVLC.SetDialogHandlers((title, text) => Task.CompletedTask,
                (dialog, title, text, username, store, token) => Task.CompletedTask,
                (dialog, title, text, type, cancelText, actionText, secondActionText, token) => Task.CompletedTask,
                (dialog, title, text, indeterminate, position, cancelText, token) => Task.CompletedTask,
                (dialog, position, text) => Task.CompletedTask);
        }

        private static void UnsetDialogHandlers(LibVLC libVlc)
        {
            libVlc.UnsetDialogHandlers();
        }
    }
}
```

PR #87, opened 2019-12-28
Centralize the static callbacks management

jeremyVignelles:
#### Rationale ####
There are a few APIs in libvlc that requires a callback to be passed:
- Event API
- Exit handler API
- Log API
- Dialogs API
- new Media from Callbacks API
- Video callbacks API
- Audio callbacks API

After implementation, we discovered that apple platforms didn't allow callbacks to be called on non-static method, and that the callback method must be decorated with `[MonoPInvokeCallback(typeof(CallbackDelegate))]`.

Since we implemented the APIs by waves, we were quite inconsistent about how we were doing:
- Some stored the native callback handle to avoid garbage collection, other didn't (see #84)
- Sometimes, there were `#if APPLE` with non-static code otherwise
- In the event manager, a proxy "Attach"/"Detach" logic was made with proxy methods that calls user methods, whereas in other places, the user code is passed directly...
- Sometimes the native callback handle is static, which means that it can't work with multiple instances of the class. Some other times, the static code is put in a static dictionary.

The resulting code could be complex, and sometimes not properly implemented.

#### Proposal ####

For each callback API, there is one and only one static method, marked with `[MonoPInvokeCallback(typeof(CallbackDelegate))]` in the `LibVLCCallbacks` class.

This method is the one that gets passed to libvlc, and it is responsible to route the callback to the appropriate method on the appropriate instance. How? Each libvlc callback API has a `void* opaque` pointer that we can use just for that. In that case, we expect the opaque pointer to be a GCHandle to the instance on which we call the method.
On the subscribing end, we register the user callback in a field, and then subscribe with the LibVLCCallbacks.StaticMethodCallback, and with the opaque pointer set to `GCHandle.ToIntPtr(ThisHandle)`

On all types that inherits `Internal`, `ThisHandle` is already defined. Otherwise, this is a handle you allocate with `GCHandle.Alloc(this)`.

There is one catch though. If I pass the static callback method directly, that wouldn't solve the callback on collected delegate issue. This is why, for each callback, there is also a `public static readonly xxxHandle = xxx;` that points to the static callback. That way, the static callback can be made private.

This should theorically work for iOS (not tested).

#### Advantages ####
- Single code path between platforms
- Simplify the code in classes like the event managers (less "callback handle" fields)
- Fixes a few implementations that didn't follow all the callbacks requirements (keep a reference, static method...)
- Allows to specify an implementation of a media with callbacks other than just streams.

#### Drawbacks ####
- There is an API change for the "new media with callback" API (see below). The new MediaInput type needs to be Dispose()'d
- Performance overhead of having one proxy method and GCHandle.FromIntPtr
- This implementation is not the most clean I've ever made, there are some internal fields accessed from LibVLCCallbacks. This can probably be improved

### Issues Resolved ### 
There are currently no ticket open for that as far as I know, but I have seen a few issues that are resolved by this PR:
- Some APIs had static "callback handles", which wouldn't work in case of multiple instances
- Some APIs wouldn't work for iOS, like the video callback API.

### API Changes ###
- SetExitHandler doesn't take a second argument (it wasn't used anyway)
- Media constructor does not accept a Stream as input, but rather a MediaInput (see below)

### Platforms Affected ### 
- Core (all platforms)

Callbacks calls might be a little slower, but apart from that and the API changes, there are no expected behavioral change.

### A word about MediaInput ###
The previous implementation let the user pass a Stream into the Media constructor.
It then allocates a `StreamData` instance that was then inserted in a static Dictionary instance.

I wanted to avoid that static Dictionary, and use my LibVLCCallbacks class instead, but which GCHandle would I pass as `opaque`? I need to keep a reference to the GCHandle, to be able to free it afterwards, but Media is not a good candidate for that, since it is disposed immediately after being used.

I didn't find a better way than having an additional MediaInput class. One can either use the StreamMediaInput, or create its own by implementing the abstract methods.

This is a breaking change, but avoids using a static dictionary.

Not applicable

### Testing Procedure ###
Unit tests (I couldn't run them all, and some are failing because of 404)

### PR Checklist ###


mfkl:
Good job dude! 🎉 

A few remarks.

> This implementation is not the most clean I've ever made, there are some internal fields accessed from LibVLCCallbacks. This can probably be improved

This is my biggest painpoint with your PR. 

As you know, I always try to stay as close to the libvlc api as possible. I think we could do without that `LibVLCCallbacks` class, notably because it requires having a few `internal` fields and methods (like the dialog callbacks) on main libvlc types. We could avoid that, by having each callback declarations in their own respective types.

The `GetInstance<T>` method is a good candidate for `MarshalUtils` IMO.

> This should theorically work for iOS (not tested).

I will.

> Media constructor does not accept a Stream as input, but rather a MediaInput (see below)

I understand why you made that change in that PR, though to me it's unrelated to the refactoring part (notably because it breaks the API). It should be fine, I guess.

> Unit tests (I couldn't run them all, and some are failing because of 404)

You think there is still a point in maintaining those?

LibVLCSharp/Shared/Events/MediaEventManager.cs
```diff
@@ -1,208 +1,98 @@
 ﻿using System;
-using LibVLCSharp.Shared.Helpers;
 
 namespace LibVLCSharp.Shared
 {
     internal class MediaEventManager : EventManager
     {
         readonly object _lock = new object();
```
mfkl: You can remove this as well I guess. You think the locks aren't needed anymore? Not sure they ever were

LibVLCSharp/Shared/Events/EventManager.cs
```diff
@@ -11,16 +10,38 @@ internal abstract class EventManager
...
+        /// </summary>
+        internal class EventTypeManager: IDisposable
+        {
+            internal Action<IntPtr> EventHandler { get; }
+            internal GCHandle _handle;
+
+            internal EventTypeManager(Action<IntPtr> eventHandler)
+            {
+                EventHandler = eventHandler;
+                _handle = GCHandle.Alloc(this);
```
mfkl: Guess you can use the same casing here since they're both `internal`

LibVLCSharp/Shared/Events/EventManager.cs
```diff
@@ -85,7 +116,4 @@ internal protected void OnEventUnhandled(object sender, EventType eventType)
         internal protected abstract void AttachEvent<T>(EventType eventType, EventHandler<T> eventHandler) where T : EventArgs;
```
mfkl: I'm guessing that with your refactoring `#if APPLE` above (can't select it in github) should not be needed anymore?

LibVLCSharp/Shared/LibVLCCallbacks.cs
```diff
@@ -0,0 +1,491 @@
...
+using System.Threading.Tasks;
+using LibVLCSharp.Shared.Helpers;
+
+namespace LibVLCSharp.Shared
+{
+    /// <summary>
+    /// A static class that contains callbacks that libvlc can call.
+    ///
+    /// In many platforms, we can use any method or lambda as a native callback,
+    /// but not on Apple platforms, which requires static methods, decorated with [MonoPInvokeCallback]
```
mfkl: Or Unity

LibVLCSharp/Shared/LibVLCCallbacks.cs
```diff
@@ -0,0 +1,491 @@
...
+        {
+            var libVLC = GetInstance<LibVLC>(libVLCHandle);
+
+            try
+            {
+                var message = MarshalUtils.GetLogMessage(format, args);
+
+                LibVLC.GetLogContext(logContext, out var module, out var file, out var line);
+
+                Action logAction = () => libVLC._log?.Invoke(null, new LogEventArgs(logLevel, message, module, file, line));
```
mfkl: You can use a local function here

` void logAction() => libVLC._log?.Invoke(null, new LogEventArgs(logLevel, message, module, file, line));`

LibVLCSharp/Shared/Events/MediaEventManager.cs
jeremyVignelles: indeed, the lock has been pushed down to the attach/detach method

LibVLCSharp/Shared/Events/EventManager.cs
jeremyVignelles: Yes, that should be a property instead of a field

mfkl:
Tested on iOS 👍

LibVLCSharp/Shared/Internal.cs
```diff
@@ -22,6 +23,11 @@ public abstract class Internal : IDisposable
         /// </summary>
         protected bool IsDisposed;
 
+        /// <summary>
+        /// The GCHandle to be passed to callbacks as userData
+        /// </summary>
+        protected GCHandle ThisHandle;
```
mfkl: Since this is only used with the `MediaPlayer` and `LibVLC` objects, maybe it could be moved down to those types? The other libvlc objects don't use it, so thats an unneeded allocation.

LibVLCSharp/Shared/Media.cs
```diff
@@ -253,10 +254,10 @@ static IntPtr CtorFromInput(LibVLC libVLC, MediaInput input)
...
             return Native.LibVLCMediaNewCallbacks(libVLC.NativeReference,
-                LibVLCCallbacks.OpenMediaCallbackHandle,
-                LibVLCCallbacks.ReadMediaCallbackHandle,
-                LibVLCCallbacks.SeekMediaCallbackHandle,
-                LibVLCCallbacks.CloseMediaCallbackHandle,
+                OpenMediaCallbackHandle,
+                ReadMediaCallbackHandle,
+                SeekMediaCallbackHandle,
+                CloseMediaCallbackHandle,
                 GCHandle.ToIntPtr(input.ThisHandle));
```
jeremyVignelles: Can we rename "ThisHandle" to something else (GCHandle maybe?)

mfkl: Done

mfkl:
Now I'm wondering if we really need the `MediaInput` abstract class...

> Override this to provide your own reading mechanism, or you can use the <see cref="StreamMediaInput"/> class

What other mechanisms can you think of?

jeremyVignelles:
Pipelines, read with an ArrayPool, native API/http API/websocket API calls to read the next blocks...

PR #90, opened 2020-01-16
Fix a memory leak with the foreground window (WPF)

srollinet:
### Description of Change ###

Fix a leak with the foreground window when the VideoView is disposed

In VideoView.Dispose()
- close the foreground window
- dispose the PlayerHost

In the WPF samples
- dispose the VideoView when the window is closed

### Issues Resolved ### 

None

### API Changes ###

 None

### Platforms Affected ### 

WPF

None

Not applicable

### Testing Procedure ###

- Start the WPF sample with a memory profiler
- Open Example 1 and close the window
- Open Example 2 and close the window
- Take a memory snapshot
- Check that no instance of ForeroundWindow or VideoView have survived

### PR Checklist ###


mfkl:
Thanks @srollinet! Will release a new version soon.

PR #92, opened 2020-01-27
Core: Always load libcpp on Android regardless of Android version

mfkl:
This was partly fixed in https://github.com/videolan/libvlcsharp/pull/88 but not entirely, according to https://code.videolan.org/videolan/LibVLCSharp/issues/283. This commit should fix it for good.

### Description of Change ###

Always load libcpp on Android regardless of Android version

### Issues Resolved ### 

- fixes https://code.videolan.org/videolan/LibVLCSharp/issues/283

### API Changes ###
 
 None

### Platforms Affected ### 

- Android


mfkl:
Waiting on update from https://code.videolan.org/videolan/LibVLCSharp/issues/283

PR #98, opened 2020-02-19
CI: Fix Linux build

mfkl:


mfkl:
Can repro locally, libvlcsharp doesn't build on linux with .NET SDK 3
```
/usr/share/dotnet/sdk/3.1.100/Sdks/Microsoft.NET.Sdk.WindowsDesktop/targets/Microsoft.NET.Sdk.WindowsDesktop.props(208,3): error MSB4019: The imported project "/usr/share/dotnet/sdk/3.1.100/Sdks/Microsoft.NET.Sdk.WindowsDesktop/targets/Microsoft.WinFX.props" was not found. Confirm that the expression in the Import declaration "Microsoft.WinFX.props" is correct, and that the file exists on disk.
```

mfkl:
Building with `2.1.700` works fine.

mfkl:
Need to try on newer Ubuntu

PR #102, opened 2020-03-07
Initial Libvlc 4 support

mfkl:
WIP

- [ ] expose a few thumbnailer missing methods 
- [ ] Adress TODO
- [ ] `bool crop` libvlc missing doc
- [ ] Fix wrong `ThumbnailerRequestByTime` return comment
- [ ] clarify destroy/cancel thumbnailing APIs
- [ ] Test all

Lots more APIs to implement/remove, this ticket https://github.com/ZeBobo5/Vlc.DotNet/issues/545 contains a good record.

Might merge this first and make multiple PRs for the rest.


mfkl:
```
2020-03-09T15:58:32.1310720Z Shared/Picture.cs(73,16): error CS0122: 'ReadOnlyMemory<T>' is inaccessible due to its protection level [/Users/runner/runners/2.165.0/work/1/s/LibVLCSharp/LibVLCSharp.csproj]
2020-03-09T15:58:32.1312650Z Shared/Picture.cs(29,36): error CS0122: 'ReadOnlyMemory<T>' is inaccessible due to its protection level [/Users/runner/runners/2.165.0/work/1/s/LibVLCSharp/LibVLCSharp.csproj]
```
But _only_ on the macOS CI

LibVLCSharp/Shared/Picture.cs
```diff
@@ -0,0 +1,127 @@
...
+        }
+
+        // TODO: test this, we probably don't want to expose UIntPtr as a param
+        /// <summary>
+        /// The image internal buffer, including potential padding.
+        /// The Picture owns the returned buffer, which must not be modified nor freed
+        /// </summary>
+        /// <param name="size">A pointer to a size_t that will hold the size of the buffer</param>
+        /// <returns>A pointer to the internal buffer.</returns>
+        public ReadOnlyMemory<byte> Buffer(UIntPtr size) => Native.LibVLCPictureGetBuffer(NativeReference, size);
```
jeremyVignelles: Not sure we can do that directly, the ReadOnlyMemory has no way to know the size.
I found this : https://github.com/dotnet/runtime/issues/29530 and that https://github.com/dotnet/corefx/blob/0d3caeb/src/Common/src/System/Memory/PointerMemoryManager.cs , but it needs to use ` unsafe` code. Too bad...

I found `SafeBuffer`, but the doc says it's deprecated, but didn't find why anywhere : https://docs.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.safebuffer?view=netcore-3.1

jeremyVignelles: Anyway, the size pointer is probably a bad idea, it should be a property of what gets returned.

PR #117, opened 2020-04-16
Fixes #294

kakone:
### Description of Change ###
Adds the LibVLCSharp project as an assembly reference instead of a project reference. With this change, the LibVLCSharp.Uno.pri file is correctly generated for the UWP platform.

`<Reference Include="$(SolutionDir)LibVLCSharp\bin\$(Configuration)\$(TargetFramework)\LibVLCSharp.dll" />`

Disadvantage : the NuGet package can't be automatically generated on build anymore (because the package reference to LibVLCSharp is missing). A new nupsec file must be created.

### Issues Resolved ### 
- fixes https://code.videolan.org/videolan/LibVLCSharp/-/issues/294

### API Changes ###
None

### DIFF
--- a/src/LibVLCSharp.sln
+++ b/src/LibVLCSharp.sln
@@ -11,7 +11,7 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Solution Items", "Solution 
 		global.json = global.json
 	EndProjectSection
 EndProject
-Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Samples", "..\samples", "{799A84A2-2161-4676-878B-5610E3586137}"
+Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Samples", "Samples", "{799A84A2-2161-4676-878B-5610E3586137}"
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "LibVLCSharp.Forms", "LibVLCSharp.Forms\LibVLCSharp.Forms.csproj", "{06814458-C8B4-4D50-86E2-904E6A9E3499}"
 EndProject
--- a/buildsystem/build.cake
+++ b/buildsystem/build.cake
@@ -14,7 +14,8 @@ var packagesDir = "../packages";
 //////////////////////////////////////////////////////////////////////
 
 // Define directories.
-var buildDir = Directory("./build") + Directory(configuration);
+var artifacts = Directory("./build") + Directory(configuration);
+var builds = "../src/**/bin/Release/*.nupkg";
 
 //////////////////////////////////////////////////////////////////////
 // TASKS
@@ -23,8 +24,8 @@ var buildDir = Directory("./build") + Directory(configuration);
 Task("Clean")
     .Does(() =>
 {
-    DeleteFiles("./**/bin/Release/*.nupkg");
-    CleanDirectory(buildDir);
+    DeleteFiles(builds);
+    CleanDirectory(artifacts);
     if(DirectoryExists(packagesDir))
     {
         DeleteDirectory(packagesDir, new DeleteDirectorySettings 
@@ -46,14 +47,14 @@ Task("Build")
     .IsDependentOn("Restore-NuGet-Packages")
     .Does(() =>
 {
-    MSBuild(solutionPath, settings => settings.SetConfiguration(configuration));
+    MSBuild(solutionPath, settings => settings.SetConfiguration(configuration));    
 });
 
 Task("CopyNugets")
     .IsDependentOn("Build")
     .Does(() =>
 {
-    CopyFiles(GetFiles("./**/bin/Release/*.nupkg"), buildDir);
+    CopyFiles(GetFiles(builds), artifacts);
 });
 
 //////////////////////////////////////////////////////////////////////
--- a/samples/Uno/LibVLCSharp.Uno.Sample.Droid/LibVLCSharp.Uno.Sample.Droid.csproj
+++ b/samples/Uno/LibVLCSharp.Uno.Sample.Droid/LibVLCSharp.Uno.Sample.Droid.csproj
@@ -65,9 +65,9 @@
     <Reference Include="System.Xml" />
   </ItemGroup>
   <ItemGroup>
-    <PackageReference Include="Uno.UI" Version="2.0.512-dev.4178" />
+    <PackageReference Include="Uno.UI" Version="2.0.527" />
     <PackageReference Include="VideoLAN.LibVLC.Android">
-      <Version>3.1.2.1</Version>
+      <Version>3.2.0</Version>
     </PackageReference>
   </ItemGroup>
   <ItemGroup>
--- a/samples/Uno/LibVLCSharp.Uno.Sample.iOS/LibVLCSharp.Uno.Sample.iOS.csproj
+++ b/samples/Uno/LibVLCSharp.Uno.Sample.iOS/LibVLCSharp.Uno.Sample.iOS.csproj
@@ -115,9 +115,9 @@
     <BundleResource Include="Resources\Fonts\winjs-symbols.ttf" />
   </ItemGroup>
   <ItemGroup>
-    <PackageReference Include="Uno.UI" Version="2.0.512-dev.4178" />
+    <PackageReference Include="Uno.UI" Version="2.0.527" />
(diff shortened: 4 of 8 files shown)


mfkl:
> Adds the LibVLCSharp project as an assembly reference instead of a project reference. With this change, the LibVLCSharp.Uno.pri file is correctly generated for the UWP platform.

Do you know whether that's a Uno problem or an MSBuild problem?

mfkl:
Thanks for tracking this down btw 👍

kakone:
> Do you know whether that's a Uno problem or an MSBuild problem?

It's an MSBuild problem. I reproduced it here with an [UWP sample project](https://github.com/kakone/PriGeneration).
I asked a [question on stackoverflow](https://stackoverflow.com/questions/60695801/uwp-class-librairies-and-pri-file-generation) but I did not get an answer.

kakone:
I also updated the Uno.UI NuGet package to version 2.1.37 in this PR. And I restored the Android sample projects (they had disappeared from the solution).

mfkl:
> It's an MSBuild problem. I reproduced it here with an UWP sample project.

Do you know where in the MSBuild pipeline it fails? Are you willing to open a bug report on their repo with your minimal repro?

> I also updated the Uno.UI NuGet package to version 2.1.37 in this PR

That's fine, I guess. Any particular reason? We try to keep the lowest possible Uno (and XF) dependency.

> And I restored the Android sample projects (they had disappeared from the solution).

https://code.videolan.org/videolan/LibVLCSharp/-/issues/249

kakone:
> That's fine, I guess. Any particular reason? We try to keep the lowest possible Uno (and XF) dependency.

We used a preview version (2.0.512-dev). OK, I'm going to update to the lowest stable version (2.0.527).

src/LibVLCSharp.sln
```diff
@@ -11,7 +11,7 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Solution Items", "Solution
 		global.json = global.json
 	EndProjectSection
 EndProject
-Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Samples", "..\samples", "{799A84A2-2161-4676-878B-5610E3586137}"
+Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Samples", "Samples", "{799A84A2-2161-4676-878B-5610E3586137}"
```
mfkl, line 14: VS generated change? I think you can revert it.

kakone: Yes, VS is always generate this change, and I think he is right. It's a solution folder, it does not correspond to a physical path.

mfkl: Weird, my VS isn't doing this. No strong feelings.

kakone:
> Do you know where in the MSBuild pipeline it fails?

It not really fails, it's a missing option for our case. When the GenerateLibraryLayout is enabled, it generates a PRI file with the XAML resources. If we wanted to make a single NuGet package with all the dlls and all the XAML resources (a LibVLCSharp.Uno package that would include the LibVLCSharp dll), it works well. The problem is when you want to do 2 separate NuGet packages. As the second PRI file (LibVLCSharp.Uno PRI file) will contain too much things, it fails.

There are 2 workaround for this problem : to include the LibVLCSharp library as a package reference (which makes sense for Visual Studio to understand what we want to do, but it's not practical in debug mode). Or to include the LibVLCSharp library as an assembly reference (but, then, you have to make your NuGet package yourself).

mfkl:
> Or to include the LibVLCSharp library as an assembly reference (but, then, you have to make your NuGet package yourself).

I guess this is the way we have to go, at least temporarily, though I'm not too happy about it :-)

> it's a missing option for our case.

Would it make sense to raise this concern on their side then? I find it hard to believe we are the only ones having this issue.

kakone:
> I guess this is the way we have to go, at least temporarily, though I'm not too happy about it :-)
> 
> Edit: Or we use the package reference option, and keep a patch for the local debug scenario.

Yes, we can let the ProjectReference to debug the sample projects, but in release mode, it depends of the build process: the LibVLCSharp NuGet package must be built and deployed before the LibVLCSharp.Uno compilation in the PackageReference case.

kakone:
> Would it make sense to raise this concern on their side then?

Yes, I take care of it.

src/LibVLCSharp.sln
kakone: It does it on my side as soon as I make a change to the .sln file (by adding a project, or updating a project build order).

mfkl:
I'm not too happy with having to use a nuspec file again, and having to tweak CI/packaging, for only 1 project, which is our least used package by far... 

I tried [various nuget configuration schemes](https://docs.microsoft.com/en-us/nuget/consume-packages/package-references-in-project-files#controlling-dependency-assets) but could still not find a way to make it work with `ProjectReference`. It must be possible though..

kakone:
I found a workaround for this problem. I added a custom **filtered.layout.resfiles** in the project. This file contains the list of the resources needed for the PRI file (you can find this file in the obj folder, I only removed the resources of the LibVLCSharp project) :
```
LibVLCSharp.Uno\LibVLCSharp.Uno.xr.xml
LibVLCSharp.Uno\Themes\Generic.xaml
LibVLCSharp.Uno\Themes\Generic.xbf
LibVLCSharp.Uno\Themes\MediaPlayerElement.xaml
LibVLCSharp.Uno\Themes\MediaPlayerElement.xbf
LibVLCSharp.Uno\Themes\PlaybackControls.xaml
LibVLCSharp.Uno\Themes\PlaybackControls.xbf
```
Then, I copy this file before the PRI file generation (before the target **_GenerateProjectPriFileCore**) by adding these lines in the LibVLCSharp.Uno.csproj :
```
<Target Name="CustomPriFileGeneration" BeforeTargets="_GenerateProjectPriFileCore">
      <Copy SourceFiles="$(ProjectDir)\filtered.layout.resfiles" DestinationFolder="$(BaseIntermediateOutputPath)\$(Configuration)\$(TargetFramework)\" />
</Target>
```

I can commit these changes if you prefer this solution.

mfkl:
> I can commit these changes if you prefer this solution.

Yes, this is the better solution IMO.

mfkl:
```
LibVLCSharp\samples\Uno\LibVLCSharp.Uno.Sample.Shared\MainPage.xaml : Xaml Internal Error error WMC9999: Type universe cannot resolve assembly: LibVLCSharp, Version=3.4.3.0, Culture=neutral, PublicKeyToken=null. [Li
bVLCSharp\samples\Uno\LibVLCSharp.Uno.Sample.UWP\LibVLCSharp.Uno.Sample.UWP.csproj]
```
When trying to build the Uno sample with a nuget built from your branch.

```
+    <PackageReference Include="LibVLCSharp.Uno">
+      <Version>3.4.3</Version>
+    </PackageReference>
     <PackageReference Include="Microsoft.NETCore.UniversalWindowsPlatform">
       <!-- 
 			If, in the same solution, you are referencing a project that uses https://github.com/onovotny/MSBuildSdkExtras,
@@ -100,13 +103,6 @@
     <Prefer32Bit>true</Prefer32Bit>
     <UseDotNetNativeToolchain>true</UseDotNetNativeToolchain>
   </PropertyGroup>
-  <ItemGroup>
-    <!-- A reference to the entire .Net Framework and Windows SDK are automatically included -->
-    <ProjectReference Include="..\..\..\src\LibVLCSharp\LibVLCSharp.csproj">
-      <Project>{d1c3b7c4-713b-46b2-b33a-e9298c819921}</Project>
-      <Name>LibVLCSharp</Name>
-    </ProjectReference>
-  </ItemGroup>
   <ItemGroup>
     <Compile Include="Properties\AssemblyInfo.cs" />
   </ItemGroup>
@@ -119,12 +115,6 @@
   <ItemGroup>
     <Content Include="Properties\Default.rd.xml" />
   </ItemGroup>
-  <ItemGroup>
-    <ProjectReference Include="..\..\..\src\LibVLCSharp.Uno\LibVLCSharp.Uno.csproj">
-      <Project>{82ab6f00-b0cc-4b88-8d09-cd95789e6c38}</Project>
-      <Name>LibVLCSharp.Uno</Name>
-    </ProjectReference>
-  </ItemGroup>
```

mfkl:
Nevermind https://github.com/videolan/libvlcsharp/pull/117#issuecomment-617731018, works after cleaning up...

mfkl:
Thank you, Stéphane.

PR #122, opened 2020-05-02
Update LibVLCSharp.csproj

kakone:
### Description of Change ###

Currently, the condition `$(TargetFramework)' == 'netstandard1.1'` in the LibVLCSharp.csproj removes Shared\MediaPlayerElement files for .NetStandard 1.1.
This PR updates this condition by `Condition="$(TargetFramework.StartsWith('netstandard'))"` to not include them in .NetStandard 2.0 too.

This PR also removes `netstandard2.1` from the target frameworks. I don't see the need to add a .netstandard2.1 target, bringing nothing more than the .netstandard2.0 target. Am I wrong ?

### DIFF
--- a/src/LibVLCSharp/LibVLCSharp.csproj
+++ b/src/LibVLCSharp/LibVLCSharp.csproj
@@ -28,7 +28,7 @@ This package also contains the views for the following platforms:
 If you need Xamarin.Forms support, see LibVLCSharp.Forms. 
 
 LibVLC needs to be installed separately, see VideoLAN.LibVLC.* packages.</Description>
-    <TargetFrameworks>netstandard2.1;netstandard2.0;netstandard1.1;net40;net471</TargetFrameworks>
+    <TargetFrameworks>netstandard2.0;netstandard1.1;net40;net471</TargetFrameworks>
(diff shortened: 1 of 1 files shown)


jeremyVignelles:
I added the netstandard 2.1 target because, as far as I know, nullable annotations are not available on .net standard 2.0 (though I don't think I tested very touroughly)
It was introduced by #75

kakone:
OK, thanks for the explanation (it's strange that it didn't complain when I compile it for the first time). 

So, the PR now only replaces `$(TargetFramework)' == 'netstandard1.1'` by `$(TargetFramework.StartsWith('netstandard'))`.

jeremyVignelles:
It doesn't necessarily complain, it's just that if you want to use the project in a netstandard 2.1 package, you won't know if a particular parameter can be null or not.

Just for my curiosity, why do we need to remove that file for netstandard builds?

kakone:
I close this PR because the .netstandard 2.0 target is used by LibVLCSharp.Forms, where MediaPlayerElement files are used.
I really have compilation problems on my machine because it does not indicate the errors...

kakone:
> Just for my curiosity, why do we need to remove that file for netstandard builds.

These files are InternalVisibleTo LibVLCSharp.Forms and LibVLCSharp.Uno, so we don't have to compile it for .netstandard target.

jeremyVignelles:
When you build in visual studio, it builds only for the selected TFM, not all, if I remember correctly.

You can choose the TFM next to the "debug"button.

PR #123, opened 2020-05-02
Few corrections in Uno MediaPlayerElement

kakone:
### Description of Change ###

- Corrects the SeekBarUnavailable state.
- Corrects the ZoomUnavailablity state.
- Updates the visibility of the playback controls after fade in / fade out (currently, only the opacity is modified).
- Prevent AspectRatio menu item from being unchecked
- Corrects FitScreen aspect ratio when the video track width and height are equals to 0 (perhaps it will fix [#270](https://code.videolan.org/videolan/LibVLCSharp/-/issues/270) but I'm not sure).
- Always keep the selected aspect ratio

### Issues Resolved ### 
- The IsSeekBarVisible property doesn't work on PlaybackControls, so the seek bar can't be hidden.
- The IsZoomButtonVisible property doesn't work

### API Changes ###
None

### Platforms Affected ### 
- UWP / Uno MediaPlayerElement

### DIFF
--- a/src/LibVLCSharp.Uno/PlaybackControlsBase.cs
+++ b/src/LibVLCSharp.Uno/PlaybackControlsBase.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Collections.Generic;
 using System.Linq;
 using FontAwesome;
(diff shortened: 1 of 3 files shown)


mfkl:
> perhaps it will fix #270 but I'm not sure

Afraid not, just tested. 

It also appears LibVLCSharp.Forms doesn't build from VS Mac (fine from CLI), but that's unrelated to your PR.

I'll merge this, thanks!

PR #129, opened 2020-05-22
Declared non-seekable streams as non-seekable in callback input

jeremyVignelles:
### Description of Change ###

These are the same changes as https://github.com/ZeBobo5/Vlc.DotNet/pull/648

Streams where CanSeek is false should not be declared as seekable in libvlc.
Also increased the buffer size to match the libvlc buffer size.

### Issues Resolved ### 

None opened on LVS, but see https://github.com/ZeBobo5/Vlc.DotNet/issues/647

### API Changes ###
Added:
 - bool MediaInput.CanSeek { get; protected set; } // Defaults to true, must be set to false in subclasses constructor.

Changed:
 - Native.LibVLCMediaNewCallbacks now has a nullable parameter for seekCb
 
### Platforms Affected ### 
- Core (all platforms)

Will not attempt to call Seek if the `Stream` is not seekable in `StreamMediaInput`. LibVLC will now understand that the stream is not seekable and will show appropriate log messages.

Not applicable

### Testing Procedure ###
Didn't have time to push it to a repo, sorry:

```cs
            Core.Initialize();

            using var libVLC = new LibVLC("-vv");

            var FileStream = new FileStream(path: @"D:\Projects\Tests\vlc_text_encr\Encrypted_Video\video.ded", mode: FileMode.Open, access: FileAccess.Read);
            RijndaelManaged AES = new RijndaelManaged();
            SHA256 SHA256 = SHA256.Create();

            // read key from TextBox4.Text
            AES.Key = SHA256.ComputeHash(Encoding.ASCII.GetBytes("U[#x5:jg0$e-^etBx#MjWH5Zu_ndd9"));
            AES.Mode = CipherMode.ECB;


            // cryptostream starts here
            var cryptoStream = new CryptoStream(FileStream, AES.CreateDecryptor(), CryptoStreamMode.Read);
            using var input = new StreamMediaInput(cryptoStream);
            using var media = new Media(libVLC, input);
            using var mp = new MediaPlayer(media);
            mp.Play();
            Console.ReadKey();
```

The video can be found here: https://github.com/graysuit/vlc_text_encr/tree/master/Encrypted_Video

Expected result : "mp4 stream error: no moov before mdat and the stream is not seekable" in the log

### PR Checklist ###

### DIFF
--- a/samples/LibVLCSharp.NetCore.Sample/LibVLCSharp.NetCore.Sample.csproj
+++ b/samples/LibVLCSharp.NetCore.Sample/LibVLCSharp.NetCore.Sample.csproj
@@ -7,7 +7,7 @@
     <LangVersion>8.0</LangVersion>
   </PropertyGroup>
   <ItemGroup>
-    <PackageReference Condition="$([MSBuild]::IsOsPlatform('Windows'))" Include="VideoLAN.LibVLC.Windows" Version="3.0.8" />
+    <PackageReference Condition="$([MSBuild]::IsOsPlatform('Windows'))" Include="VideoLAN.LibVLC.Windows" Version="3.0.11" />
     <PackageReference Condition="$([MSBuild]::IsOsPlatform('OSX'))" Include="VideoLAN.LibVLC.Mac" Version="3.1.3" />
   </ItemGroup>
   <ItemGroup>
--- a/src/LibVLCSharp/Shared/Media.cs
+++ b/src/LibVLCSharp/Shared/Media.cs
@@ -44,7 +44,7 @@ namespace LibVLCSharp.Shared
             [DllImport(Constants.LibraryName, CallingConvention = CallingConvention.Cdecl,
                 EntryPoint = "libvlc_media_new_callbacks")]
             internal static extern IntPtr LibVLCMediaNewCallbacks(IntPtr libVLC, InternalOpenMedia openCb, InternalReadMedia readCb, 
-                InternalSeekMedia seekCb, InternalCloseMedia closeCb, IntPtr opaque);
+                InternalSeekMedia? seekCb, InternalCloseMedia closeCb, IntPtr opaque);
 
             [DllImport(Constants.LibraryName, CallingConvention = CallingConvention.Cdecl,
                 EntryPoint = "libvlc_media_add_option")]
@@ -284,7 +284,7 @@ namespace LibVLCSharp.Shared
             return Native.LibVLCMediaNewCallbacks(libVLC.NativeReference,
                 OpenMediaCallbackHandle,
                 ReadMediaCallbackHandle,
-                SeekMediaCallbackHandle,
+                input.CanSeek ? SeekMediaCallbackHandle : null,
                 CloseMediaCallbackHandle,
                 GCHandle.ToIntPtr(input.GcHandle));
         }
--- a/src/LibVLCSharp/Shared/MediaInput.cs
+++ b/src/LibVLCSharp/Shared/MediaInput.cs
@@ -24,6 +24,11 @@ namespace LibVLCSharp.Shared
         }
 
         /// <summary>
+        /// A value indicating whether this Media input can be seeked in.
+        /// </summary>
+        public bool CanSeek { get; protected set; } = true;
+
+        /// <summary>
         /// LibVLC calls this method when it wants to open the media
         /// </summary>
         /// <param name="size">This value must be filled with the length of the media (or ulong.MaxValue if unknown)</param>
--- a/src/LibVLCSharp/Shared/MediaInput.cs
+++ b/src/LibVLCSharp/Shared/MediaInput.cs
@@ -40,7 +40,7 @@ namespace LibVLCSharp.Shared
         /// </summary>
         /// <param name="buf">The buffer where read data must be written</param>
         /// <param name="len">The buffer length</param>
-        /// <returns>The number of bytes actually read, -1 on error</returns>
+        /// <returns>strictly positive number of bytes read, 0 on end-of-stream, or -1 on non-recoverable error</returns>
         public abstract int Read(IntPtr buf, uint len);
 
         /// <summary>
--- a/src/LibVLCSharp/Shared/StreamMediaInput.cs
+++ b/src/LibVLCSharp/Shared/StreamMediaInput.cs
@@ -24,6 +24,7 @@ namespace LibVLCSharp.Shared
         public StreamMediaInput(Stream stream)
         {
             _stream = stream ?? throw new ArgumentNullException(nameof(stream));
+            CanSeek = stream.CanSeek;
         }
 
         /// <summary>
--- a/src/LibVLCSharp/Shared/StreamMediaInput.cs
+++ b/src/LibVLCSharp/Shared/StreamMediaInput.cs
@@ -65,7 +65,7 @@ namespace LibVLCSharp.Shared
         /// </summary>
         /// <param name="buf">The buffer where read data must be written</param>
         /// <param name="len">The buffer length</param>
-        /// <returns>The number of bytes actually read, -1 on error</returns>
+        /// <returns>strictly positive number of bytes read, 0 on end-of-stream, or -1 on non-recoverable error</returns>
         public unsafe override int Read(IntPtr buf, uint len)
         {
             try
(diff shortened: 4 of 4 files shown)


mfkl:
> Also increased the buffer size to match the libvlc buffer size.

Is this buffer size documented anywhere in the libvlc docs?

jeremyVignelles:
Indeed not, not as I'm aware of. It's probably something that depend on the access.
I just tried to debug and found that libvlc was asking me a buffer that big. I didn't try to dig and find why.

I'm ok to revert that change if needed.

mfkl:
Hmm something's off with the `CanSeek` property

```
using System;
using System.Net.Http;
using System.Threading.Tasks;
using LibVLCSharp.Shared;

namespace LibVLCSharp.NetCore.Sample
{
    class Program
    {
        static async Task Main(string[] args)
        {
            Core.Initialize();

            using var libVLC = new LibVLC("-vv");

            var client = new HttpClient();
            var stream = await client.GetStreamAsync("http://commondatastorage.googleapis.com/gtv-videos-bucket/sample/ElephantsDream.mp4");

            using var input = new StreamMediaInput(stream);
            using var media = new Media(libVLC, input);
            using var mp = new MediaPlayer(media);
            mp.Play();
            Console.ReadKey();
        }
    }
}
```

This doesn't work on your branch. Fine with if I comment out https://github.com/videolan/libvlcsharp/blob/dfb9644a6a2aede41d5c9bcd552e74b72580776a/src/LibVLCSharp/Shared/StreamMediaInput.cs#L23

mfkl:
rebased and updated to latest libvlc. Also re-added the return 0 for non seekable stream. I can't repro the problem with my repro from https://github.com/videolan/libvlcsharp/pull/129#issuecomment-633538785

mfkl:
> I just tried to debug and found that libvlc was asking me a buffer that big.

Do you recall the log message?

mfkl:
```
main Debug: no stream_filter modules matched
main Debug: looking for stream_directory module matching "any": 1 candidates
main Debug: no stream_directory modules matched
main Debug: looking for stream_filter module matching "record": 24 candidates
main Debug: using stream_filter module "record"
main Debug: creating demux: access='imem' demux='any' location='' file='(null)'
main Debug: looking for demux module matching "any": 55 candidates
main Debug: attachment of directory-extractor failed for imem://
mp4 Warning: truncated box mdat discarded
mp4 Warning: truncated box mdat discarded
mp4 Warning: MP4 plugin discarded (not a valid initialization chunk)
mp4 Debug: freeing all memory
main Debug: looking for xml reader module matching "any": 1 candidates
main Debug: using xml reader module "xml"
mp4 Error: no moov before mdat and the stream is not seekable
webvtt Debug: subtitle demux discarded
ts Debug: TS module discarded (lost sync)
mod Debug: MOD validation failed (ext=)
avcodec Debug: CPU flags: 0x0007d3db
avcodec Debug: detected format: mov,mp4,m4a,3gp,3g2,mj2
avcodec Debug: trying url: imem://
prefetch Error: cannot seek (to offset 48)
prefetch Debug: end of stream
main Debug: selecting program id=0
avcodec Debug: adding es: audio codec = mp4a (86018)
avcodec Debug: AVFormat(ffmpeg Lavf58.3.100) supported stream
avcodec Debug:     - start time = -1
avcodec Debug:     - duration = 117312000
main Debug: using demux module "avcodec"
main Debug: looking for video decoder module matching "any": 19 candidates
avcodec Debug: using ffmpeg Lavc58.6.103
avcodec Debug: adding es: video codec = h264 (27)
avcodec Debug: CPU flags: 0x0007d3db
avcodec Debug: allowing 5 thread(s) for decoding
avcodec Debug: using frame thread mode with 5 threads
avcodec Debug: codec (h264) started
main Debug: looking for audio decoder module matching "any": 23 candidates
main Debug: using video decoder module "avcodec"
main Debug: using audio decoder module "faad"
main Debug: looking for meta reader module matching "any": 2 candidates
main Debug: no meta reader modules matched
main Debug: `imem://' successfully opened
The thread 0x4698 has exited with code 0 (0x0).
main Debug: EOF reached
The thread 0x13c0 has exited with code 0 (0x0).
main Debug: killing decoder fourcc `h264'
main Debug: removing module "avcodec"
main Debug: killing decoder fourcc `mp4a'
main Debug: removing module "faad"
main Debug: removing module "avcodec"
main Debug: removing module "record"
main Debug: removing module "prefetch"
main Debug: removing module "imem_access"
avcodec Debug:     - format = mov,mp4,m4a,3gp,3g2,mj2 (QuickTime / MOV)
main Debug: Program doesn't contain anymore ES
The thread 0x2214 has exited with code 0 (0x0).
The thread 0x439c has exited with code 0 (0x0).
The thread 0x229c has exited with code 0 (0x0).
```
That's VS logs. In the console log window (we should really find a way to pipe the log output correctly), there is:

```
[mov,mp4,m4a,3gp,3g2,mj2 @ 000001e86d1a7fe0] Protocol name not provided, cannot determine if input is local or a network protocol, buffers and access patterns cannot be configured optimally without knowing the protocol
Invalid return value 0 for stream protocol                                                                              
Invalid return value 0 for stream protocol                                                                              
[mov,mp4,m4a,3gp,3g2,mj2 @ 000001e86d1a7fe0] Could not find codec parameters for stream 0 (Video: h264, 1 reference frame (avc1 / 0x31637661), none(left), 1280x720, 1048 kb/s): unspecified pixel format                                       
Consider increasing the value for the 'analyzeduration' and 'probesize' options                                         [mov,mp4,m4a,3gp,3g2,mj2 @ 000001e86d1a7fe0] stream 0, offset 0x3f7: partial file  
```

Code: https://github.com/mfkl/libvlcsharp/tree/fix/stream-seek-repro

mfkl:
Do you want to have another look at this @jeremyVignelles?

jeremyVignelles:
I tried the testing procedure and got the "expected result". Could you remember me what else should I look for ?

mfkl:
> 
> 
> I tried the testing procedure and got the "expected result". Could you remember me what else should I look for ?

I'll have another look and test.

mfkl:
Rebased.

mfkl:
So,

```
var client = new HttpClient();
var stream = await client.GetStreamAsync("http://commondatastorage.googleapis.com/gtv-videos-bucket/sample/BigBuckBunny.mp4");
```
In that case the returned stream concrete class is a `System.Net.Http.HttpConnection.ContentLengthReadStream` which is not seekable.

Likewise, the dotnet crypto stream is not seekable by default https://docs.microsoft.com/en-us/dotnet/api/system.security.cryptography.cryptostream.canseek?view=netcore-3.1#System_Security_Cryptography_CryptoStream_CanSeek

So I'm not sure we can trust dotnet's stream implementations regarding the seeking status? It might not always fit our underlying streams capabilities.

jeremyVignelles:
Not sure I'm understanding correctly here. Why would a non-seekable stream be declared as seekable, from a logical standpoint?

It seems to work with the current implementation because the plugin prober is correct on first try, and that there is no need to seek, right? what guarantee do we have that the plugins are loaded in the same order on every platform? Will the order remain the same between libvlc versions?

At that point, the argument is that it might break the users that rely on the current behavior (working without needing to seek), and maybe we should find an alternative for them to make the transition as smooth as possible, but I think that correctness should be favored here.

Alternatives might include creating a wrapper in the same way as my SeekableStreamWrapper, but without restarting the stream every time. I'm thinking about something that would put the data from the client in a MemoryStream buffer or something, but that would block the Read until there is more data available, or the real end of stream.

mfkl:
Note: What happens when `MediaPlayerSeekableChanged` gets fired? A non-seekable stream becomes seekable or the other way around?

jeremyVignelles:
How could that happen? we declare the input as either seekable or non seekable, there's no reason that could change over time, and even then, how would that impact the input stream?

mfkl:
> How could that happen? 

https://github.com/videolan/vlc-3.0/blob/738da6c2737d2da93b24b7cce47944d45d10d8d3/lib/media_player.c#L1002

> how would that impact the input stream?

Would the .NET stream become seekable mid-stream if the underlying libvlc stream becomes so?

jeremyVignelles:
Isn't it the other way around?
How should libvlc know if the underlying .net stream became seekable? I don't think it would know, but I don't think that .net streams magically change their CanSeek property on the fly.

One thing that might be possible is that the stream is seekable, vlc starts playing it, and then realizes that it can't seek inside and raises this event? I really don't know what this is used for.

mfkl:
Just asked, seems the seekable changed event is _probably_ not fired mid-stream. Will merge this then.

mfkl:
Thanks :-) (sorry took a while to merge)

PR #135, opened 2020-06-04
build: centralize version for all projects

mfkl:


src/Directory.Build.props
```diff
@@ -17,4 +17,7 @@
   <PropertyGroup Condition="'$(TargetFramework)' == 'uap10.0'">
     <DefineConstants>$(DefineConstants);UWP10_0</DefineConstants>
   </PropertyGroup>
+  <PropertyGroup>
+    <Version>3.4.5</Version>
```
jeremyVignelles, line 21: Would it be possible to use `<VersionPrefix>` instead, so that we can build pre-release builds with
`dotnet pack /p:VersionSuffix=beta01`.
The .NET sdk automatically combines VersionPrefix and VersionSuffix to create the `Version` property, and it worked with MSBuild.Sdk.Extra last time I checked on net framework. Don't know about Xamarin targets though.

mfkl: We can do this and make sure it works when we will add continuous delivery of pre-release builds (soon). For now this PR just puts this all in one place, and makes it easier for current stable releases.

PR #143, opened 2020-07-07
libvlc loading rework

mfkl:
### Description of Change ###

Refactor Core.cs libvlc loading. No functional change.

I've split the work in 2 PRs. Next PR handles https://github.com/mfkl/libvlc-nuget/issues/11

### Issues Resolved ### 

- fixes https://code.videolan.org/videolan/LibVLCSharp/-/issues/355

### API Changes ###

 None

### Platforms Affected ### 

- Core (all platforms)


src/LibVLCSharp/Shared/Core/Core.Apple.cs
```diff
@@ -0,0 +1,29 @@
+﻿using System;
+
+namespace LibVLCSharp.Shared
+{
+#if APPLE
```
jeremyVignelles, line 5: Why isn't it at the beginning of the file, as the other files?

mfkl: Because that branch originally included the macOS rework, and there were different code paths for iOS/tvOS and macOS.

Fixed, that will come later.

PR #148, opened 2020-07-28
Fix tests and add them to CI

mfkl:
### Description of Change ###

Fix tests
Add unit test run as part of CI. Ultimately the goal is to add some checks on libvlc (like detecting some public API changes, callbacks not working anymore, crashes, events not firing, call hanging...) builds directly from the nightly of the day before (or something).

todo: commits need a bit of cleanup.

### Issues Resolved ### 

none in particular.

### API Changes ###

 None

### Platforms Affected ### 

- Core (all platforms)

None

Not applicable

### DIFF
--- a/buildsystem/azure-pipelines.yml
+++ b/buildsystem/azure-pipelines.yml
@@ -10,7 +10,7 @@ stages:
 - stage: Build
   variables:
     DOTNET_SKIP_FIRST_TIME_EXPERIENCE: true
-
+    DOTNET_CLI_TELEMETRY_OPTOUT: true
   jobs:
   - job: Linux
     pool:
@@ -33,6 +33,9 @@ stages:
 - stage: Test
   dependsOn: Build
   condition: succeeded('Build')
+  variables:
+    DOTNET_SKIP_FIRST_TIME_EXPERIENCE: true
+    DOTNET_CLI_TELEMETRY_OPTOUT: true
 
   jobs:
     - job: test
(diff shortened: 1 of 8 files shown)


mfkl:
https://github.com/videolan/libvlcsharp/commit/18611f09cf789b64d05e1ae4f3c6868fcbb7a1ea this breaks media.Parse, method never returns. Fine with the debugger. I think the reverse callback may be executed on the wrong context, never setting the Result on the tcs properly, thus the task not returning. However it should timeout but it doesn't...

mfkl:
@kakone ^ any idea?

kakone:
I didn't reproduce the problem. I created a Media object and call `await media.Parse();`, it works well in the two cases (with async/await or returning the Task directly).
In which case you have the problem ? Did you try with [`TaskCreationOptions.RunContinuationsAsynchronously`](https://github.com/davidfowl/AspNetCoreDiagnosticScenarios/blob/master/AsyncGuidance.md#always-create-taskcompletionsourcet-with-taskcreationoptionsruncontinuationsasynchronously) in the `TaskCompletionSource` constructor ?

However, I think it was finally a bad idea to return the Task directly (it's better from a performance point of view, but it is negligible in our case) : cf [Async guidance](https://github.com/davidfowl/AspNetCoreDiagnosticScenarios/blob/master/AsyncGuidance.md#prefer-asyncawait-over-directly-returning-task).

mfkl:
> I didn't reproduce the problem. I created a Media object and call await media.Parse();

I assume you have tested without https://github.com/videolan/libvlcsharp/commit/18611f09cf789b64d05e1ae4f3c6868fcbb7a1ea? Because with this commit, either in samples or in tests, the parse method never returns to the caller.

> Did you try with TaskCreationOptions.RunContinuationsAsynchronously in the TaskCompletionSource constructor ?

Yes, did not change anything.

> However, I think it was finally a bad idea to return the Task directly (it's better from a performance point of view, but it is negligible in our case) : cf Async guidance.

Sigh... this goes against all their documentation snippets and the note at the end is even more confusing.

mfkl:
Reverted to previous version. 

Investigated https://stackoverflow.com/questions/46058847/taskcompletionsource-with-native-async-callback-not-working-in-c-sharp amongst other things but that's not it. I believe there is a threading issue which is exposed when using https://github.com/videolan/libvlcsharp/commit/18611f09cf789b64d05e1ae4f3c6868fcbb7a1ea that prevents the native callback from setting the tcs to its final state. 

Somewhat related https://code.videolan.org/videolan/LibVLCSharp/-/issues/122

I'll stop investigating for now, no time and benefits aren't clear according to your link.

mfkl:
merged

PR #150, opened 2020-07-30
Add CI test run to libvlcsharp 4

mfkl:
just checking the build


mfkl:
some of the failing tests are libvlc 4 regressions. Ideally, we'd need a test for each libvlc callback.

mfkl:
Need to have another look, this would be useful about now..

lawchingman (approved):

koushikcs562:
thanks for the guidance! I’ll come up with a PR shortly. I understand it might be a bit challenging, but I see this as a great learning opportunity to improve the libvlcsharp CI and test reliability. Looking forward to contributing and getting feedback!

PR #154, opened 2020-08-16
Fixed some android build errors

jeremyVignelles:
Seems to fix some of the build issues, but I'm now getting namespace conflicts in the MediaPlayerElement sample project.

PR #155, opened 2020-08-16
Avoid a copy in StreamMediaInput on modern platform

jeremyVignelles:
### Description of Change ###
By leveraging Span<byte> and unsafe construct, we can get rid of a Marshal.Copy that bothered me from the beginning of the implementation of StreamMediaInput.

This implementation should give better performances around media input callback, and use less ram, but I didn't take time to benchmark.

Note : This PR is on top of #154 because I needed to build the projects... Feel free to rebase after #154 is merged

### Issues Resolved ### 

- fixes https://code.videolan.org/videolan/LibVLCSharp/-/issues/345

### API Changes ###

 None

### Platforms Affected ### 
- netstandard2.1
- Android
- Apple

(looks weird to me that the mono projects build with that change, but either they did, or I fail to build them properly. Will adapt depending on the CI result)

None

Not applicable

### Testing Procedure ###

Try using the StreamMediaInput and see if it still work.

### PR Checklist ###

### DIFF
--- a/src/LibVLCSharp/Helpers/MarshalExtensions.cs
+++ b/src/LibVLCSharp/Helpers/MarshalExtensions.cs
@@ -166,5 +166,31 @@ namespace LibVLCSharp.Helpers
                 MarshalUtils.LibVLCFree(ref nativeString);
             return Encoding.UTF8.GetString(buffer, 0, buffer.Length);
         }
+
+#if !APPLE && !ANDROID && !NETSTANDARD2_1
+        /// <summary>
+        /// The Span-based APIs on Stream are not available on older targets. Span can be backported on older TFMs through the System.Memory package,
+        /// but System.IO does not provide the same benefit. This code is extracted from dotnet/runtime to allow efficient media callbacks implementation.
+        /// </summary>
+        /// <remarks>https://github.com/dotnet/runtime/blob/c4b9dabec8186a0d61f0cc3ea0b7efea579bf24e/src/libraries/System.Private.CoreLib/src/System/IO/Stream.cs#L720-L734</remarks>
+        /// <param name="stream">the .NET stream</param>
+        /// <param name="buffer">the buffer to read</param>
+        /// <returns>number of bytes read</returns>
+        internal static int Read(this System.IO.Stream stream, Span<byte> buffer)
+        {
+            var sharedBuffer = System.Buffers.ArrayPool<byte>.Shared.Rent(buffer.Length);
+            try
+            {
+                var numRead = stream.Read(sharedBuffer, 0, buffer.Length);
+                if ((uint)numRead > (uint)buffer.Length)
+                {
+                    throw new System.IO.IOException("StreamTooLong");
+                }
+                new Span<byte>(sharedBuffer, 0, numRead).CopyTo(buffer);
+                return numRead;
+            }
+            finally { System.Buffers.ArrayPool<byte>.Shared.Return(sharedBuffer); }
+        }
+#endif
     }
 }
--- a/src/LibVLCSharp/StreamMediaInput.cs
+++ b/src/LibVLCSharp/StreamMediaInput.cs
@@ -1,7 +1,9 @@
 ﻿using System;
 using System.IO;
 using System.Runtime.InteropServices;
-
+#if !APPLE && !ANDROID && !NETSTANDARD2_1
+using LibVLCSharp.Helpers;
+#endif
 namespace LibVLCSharp
 {
     /// <summary>
@@ -10,7 +12,6 @@ namespace LibVLCSharp
     public class StreamMediaInput : MediaInput
     {
         private readonly Stream _stream;
-        private readonly byte[] _readBuffer = new byte[0x4000];
 
         /// <summary>
         /// Initializes a new instance of <see cref="StreamMediaInput"/>, which reads from the given .NET stream.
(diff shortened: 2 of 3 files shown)


mfkl:
Would have been nice to have benchmarks first to have a baseline, to have an idea of how much of an improvement we got, before (re)introducing unsafe code. Will review.

jeremyVignelles:
I thought about adding a benchmark, but how would I test it without VLC?

I came to the conclusion that the benchmark would be an external project where both code would be copy/pasted and compared, independently of the libvlc environment.
As it then became out of scope of the PR, my lazyness decided that it was not the first thing to do, because it could be done afterwards.

mfkl:
> because it could be done afterwards.

My point is that it should be done before, not after, introducing unsafe code that aims to improve perf. But whatever.

> how would I test it without VLC?

I have a few ideas, will comment on gitlab with plans.

src/LibVLCSharp/StreamMediaInput.cs
```diff
@@ -65,10 +67,16 @@ public override int Read(IntPtr buf, uint len)
         {
             try
             {
+#if NETSTANDARD2_1 || ANDROID || APPLE
```
mfkl, line 70: would be nice to have only 1 code path here, but I guess this won't be possible

jeremyVignelles: I thought about having a HAVE_SPAN that could be declared in the Directory.Build.props

jeremyVignelles:
I agree with you we should benchmark before. I posted this PR as a WIP to show my idea, not that we should merge it immediately.

mfkl:
Been looking at removing the copy for the other TFMs but it is not easy. relevant: https://stackoverflow.com/questions/1367350/how-can-i-in-c-sharp-stream-read-into-unmanaged-memory-stream?lq=1

Reading from managed memory and writing to native is not so well supported as the reverse. (not working: https://github.com/mfkl/libvlcsharp/commit/a9c094db6dae26cd23ad1a7acb17e844d35461d6). We might still be able to use the `UnmanagedMemoryStream` though.

mfkl:
I've also started looking into benchmarkdotnet. Ideally we would benchmark the `StreamMediaInput` methods (I know that'd include the marshalling cost, but that's a first step anyway), however due to how benchmarkdotnet is built, it seems you can only benchmark direct calls, i.e. `foo.bar()`. Will keep digging.

mfkl:
https://github.com/mfkl/libvlcsharp/commit/23c2bfa049a05fcf765d974c8aecc22bb3f098b4

mfkl:
https://github.com/videolan/libvlcsharp/pull/129 related

src/LibVLCSharp/Helpers/MarshalExtensions.cs
```diff
@@ -166,5 +166,29 @@ internal static IntPtr ToUtf8(this string? str)
...
         }
+
+#if !APPLE
+        /// <summary>
+        /// https://github.com/dotnet/runtime/blob/c4b9dabec8186a0d61f0cc3ea0b7efea579bf24e/src/libraries/System.Private.CoreLib/src/System/IO/Stream.cs#L720-L734
+        /// </summary>
+        /// <param name="stream"></param>
+        /// <param name="buffer"></param>
+        /// <returns></returns>
+        internal static int Read(this System.IO.Stream stream, Span<byte> buffer)
```
jeremyVignelles, line 177: How does it relates to the `stream.Read(Span<byte>)` method, and how come does it not cause any conflict for platforms where it's available ? We should really use the instance method where available, because it's optimized for some special cases (doesn't need the ArrayPool allocation) : See [here](https://github.com/dotnet/runtime/blob/c4b9dabec8186a0d61f0cc3ea0b7efea579bf24e/src/libraries/System.Private.CoreLib/src/System/IO/MemoryStream.cs#L367) and [there](https://github.com/dotnet/runtime/blob/c4b9dabec8186a0d61f0cc3ea0b7efea579bf24e/src/libraries/System.Private.CoreLib/src/System/IO/FileStream.Windows.cs#L412)

jeremyVignelles: Maybe the instance method is not yet available in the TFM we target? Would you be okay to add higher TFMs ? (like netstandard2.1 ? : https://docs.microsoft.com/en-us/dotnet/api/system.io.stream.read?view=netstandard-2.1 )

mfkl: https://github.com/videolan/libvlcsharp/blob/25dce8e43f0cb6568ca6b363fb4279e9ff0c62ed/src/LibVLCSharp/LibVLCSharp.csproj#L31 we already target 2.1

mfkl: This extension method was added because that signature was not available in some cases (on the Stream type, Span is available everywhere thanks to the nuget package). I'll modify it to have it call the regular implementation when possible.

src/LibVLCSharp/Helpers/MarshalExtensions.cs
```diff
@@ -166,5 +166,31 @@ internal static IntPtr ToUtf8(this string? str)
                 MarshalUtils.LibVLCFree(ref nativeString);
             return Encoding.UTF8.GetString(buffer, 0, buffer.Length);
         }
+
+#if !APPLE && !ANDROID && !NETSTANDARD2_1
```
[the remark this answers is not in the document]
mfkl: To me this is clear enough. Xamarin platforms + netstandard2.1 have it, the rest don't.

PR #165, opened 2020-09-16
core: Add logging helper to libvlc constructor

mfkl:
### Description of Change ###

The goal is have it enabled by default (?).
So many times people are not aware of the --verbose option, because exposing only an array of strings
means you need to go check the vlc cli page for information.

Having an enum explicitly for this makes it discoverable from the VS intellisense. It may be good to have it enabled by default.

### Issues Resolved ### 

- None, but in most I have to point to docs on how to enable logging.

### API Changes ###

Added:
 - LibVLCLog enum. Probably to be enabled by default.

### Platforms Affected ### 

- Core (all platforms)

### DIFF
--- a/src/LibVLCSharp/Shared/LibVLC.cs
+++ b/src/LibVLCSharp/Shared/LibVLC.cs
@@ -232,12 +232,62 @@ namespace LibVLCSharp.Shared
         }
 
         /// <summary>
+        /// Create and initialize a libvlc instance.
+        /// This functions accept a list of &quot;command line&quot; arguments similar to the
+        /// main(). These arguments affect the LibVLC instance default configuration.
+        /// LibVLC may create threads. Therefore, any thread-unsafe process
+        /// initialization must be performed before calling libvlc_new(). In particular
+        /// and where applicable:
+        /// <para>- setlocale() and textdomain(),</para>
+        /// <para>- setenv(), unsetenv() and putenv(),</para>
+        /// <para>- with the X11 display system, XInitThreads()</para>
+        /// (see also libvlc_media_player_set_xwindow()) and
+        /// <para>- on Microsoft Windows, SetErrorMode().</para>
+        /// <para>- sigprocmask() shall never be invoked; pthread_sigmask() can be used.</para>
+        /// On POSIX systems, the SIGCHLD signalmust notbe ignored, i.e. the
+        /// signal handler must set to SIG_DFL or a function pointer, not SIG_IGN.
+        /// Also while LibVLC is active, the wait() function shall not be called, and
+        /// any call to waitpid() shall use a strictly positive value for the first
+        /// parameter (i.e. the PID). Failure to follow those rules may lead to a
+        /// deadlock or a busy loop.
+        /// Also on POSIX systems, it is recommended that the SIGPIPE signal be blocked,
+        /// even if it is not, in principles, necessary, e.g.:
+        /// On Microsoft Windows Vista/2008, the process error mode
+        /// SEM_FAILCRITICALERRORS flagmustbe set before using LibVLC.
+        /// On later versions, that is optional and unnecessary.
+        /// Also on Microsoft Windows (Vista and any later version), setting the default
+        /// DLL directories to SYSTEM32 exclusively is strongly recommended for
+        /// security reasons:
+        /// Arguments are meant to be passed from the command line to LibVLC, just like
+        /// VLC media player does. The list of valid arguments depends on the LibVLC
+        /// version, the operating system and platform, and set of available LibVLC
+        /// plugins. Invalid or unsupported arguments will cause the function to fail
+        /// (i.e. return NULL). Also, some arguments may alter the behaviour or
+        /// otherwise interfere with other LibVLC functions.
+        /// There is absolutely no warranty or promise of forward, backward and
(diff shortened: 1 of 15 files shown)


src/LibVLCSharp/Shared/LibVLC.cs
```diff
@@ -231,13 +231,76 @@ public LibVLC(params string[] options)
...
         /// </summary>
         /// <param name="options">The options given by the user</param>
+        /// <param name="logging">logging configuration, if any</param>
         /// <returns>The patched options</returns>
-        static string[] PatchOptions(string[] options)
+        static string[] PatchOptions(string[] options, LibVLCLog? logging = null)
         {
+            if(logging != null)
+            {
+                switch (logging)
```
jeremyVignelles, line 289:
```cs
if(logging.HasValue && logging != LibVLCLog.Default)
{
   options = options.Concat(new[] { $"--verbose={(int)logging}" }).ToArray();
}
```

jeremyVignelles (commented):
Not sure about the shape of the API. Why isn't it a method in your options builder ?

mfkl:
Hmmm we already have a log enum in LVS
```
/// <summary>Logging messages level.</summary>
    /// <remarks>Future LibVLC versions may define new levels.</remarks>
    public enum LogLevel
    {
        /// <summary>Debug message</summary>
        Debug = 0,
        /// <summary>Important informational message</summary>
        Notice = 2,
        /// <summary>Warning (potential error) message</summary>
        Warning = 3,
        /// <summary>Error message</summary>
        Error = 4
    }
```

Which would correspond to 
```
enum libvlc_log_level
{
    LIBVLC_DEBUG=0,   /**< Debug message */
    LIBVLC_NOTICE=2,  /**< Important informational message */
    LIBVLC_WARNING=3, /**< Warning (potential error) message */
    LIBVLC_ERROR=4    /**< Error message */
};
```

But on their side, there is...
```
/** Message types */
enum vlc_log_type
{
    VLC_MSG_INFO=0, /**< Important information */
    VLC_MSG_ERR,    /**< Error */
    VLC_MSG_WARN,   /**< Warning */
    VLC_MSG_DBG,    /**< Debug */
};
```

And finally the VLC CLI docs, since we would enable this through libvlc_new args

```
-v, --verbose=<integer>        Verbosity (0,1,2)
          This is the verbosity level (0=only errors and standard messages,
          1=warnings, 2=debug).
```

mfkl:
Trying a new approach, just using a bool for simplicity. More log granularity can still be obtained by using the regular string options but:
- I rarely use anything else than `--verbose=2`,
- the libvlc log enum != vlc internal log enum != VLC CLI log args values. Until this is fixed, I prefer a simple API for most use cases.

Whenever we integrate with the microsoft logging extensions, this will be relevant again https://docs.microsoft.com/en-us/dotnet/api/microsoft.extensions.logging.loglevel

src/LibVLCSharp/Shared/LibVLC.cs
```diff
@@ -231,13 +231,63 @@ public LibVLC(params string[] options)
...
+        /// plugins. Invalid or unsupported arguments will cause the function to fail
+        /// (i.e. return NULL). Also, some arguments may alter the behaviour or
+        /// otherwise interfere with other LibVLC functions.
+        /// There is absolutely no warranty or promise of forward, backward and
+        /// cross-platform compatibility with regards to libvlc_new() arguments.
+        /// We recommend that you do not use them, other than when debugging.
+        /// </summary>
+        /// <param name="enableDebugLogs">enable verbose debug logs</param>
+        /// <param name="options">list of arguments (should be NULL)</param>
+        public LibVLC(bool enableDebugLogs = true, params string[] options)
```
jeremyVignelles, line 273: why `true` as a default value? How does the compiler choose when calling `new LibVLC()` ? Could you make the bool option required here?
Could you make the other constructor call this one? `this(false, options)`

jeremyVignelles (commented):
I'm OK with the bool option.
Apart from the comment, I'm OK for merging.

src/LibVLCSharp/Shared/LibVLC.cs
mfkl: > why true as a default value? 

Because the idea is to have debug logging enabled by default so that users have clues immediately on what may be going wrong. And they would have to disable it explicitly for release mode (or not?). This is debatable... as mentioned in the first line of https://github.com/videolan/libvlcsharp/pull/165#issue-488120016

> How does the compiler choose when calling new LibVLC() ? How does the compiler choose when calling new LibVLC() ? Could you make the bool option required here? Could you make the other constructor call this one? this(false, options)

No need apparently, according to https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/named-and-optional-arguments#overload-resolution

> If two candidates are judged to be equally good, preference goes to a candidate that doesn't have optional parameters for which arguments were omitted in the call. Overload resolution generally prefers candidates that have fewer parameters.

jeremyVignelles: So, it means that the second constructor is only called when a boolean is explicitely given, right?

We could then remove the = true with the same effect, right?

mfkl: Yeah, that's right. Will remove it.

PR #166, opened 2020-09-20
Added NativeReference checks where applicable

jeremyVignelles:
### Description of Change ###

Avoids potential use-after-free (after Dispose()).
Fixed usage of the WPF sample to dispose things properly (and avoid race conditions such as https://code.videolan.org/videolan/LibVLCSharp/-/issues/398 ).

### Issues Resolved ### 
- fixes #398

### API Changes ###

None

### Platforms Affected ### 

- Core (all platforms)

Might throw or return "default" results when using the API after the object has been disposed. This implementation tries not to throw when possible. Such case were already causing hard crash, so that's not a breaking change.

Not applicable

### Testing Procedure ###

Run the WPF sample (example 2), and see that it doesn't crash.

### PR Checklist ###

NOTE : There are places where the NativeReference is not guarded, mainly in the constructor's "create" callback function. Don't know what the behavior should be there...

### DIFF
--- a/samples/LibVLCSharp.WPF.Sample/Controls.xaml.cs
+++ b/samples/LibVLCSharp.WPF.Sample/Controls.xaml.cs
@@ -19,6 +19,7 @@ namespace LibVLCSharp.WPF.Sample
 
             // we need the VideoView to be fully loaded before setting a MediaPlayer on it.
             parent.VideoView.Loaded += VideoView_Loaded;
+            parent.VideoView.Unloaded += VideoView_Unloaded;
             PlayButton.Click += PlayButton_Click;
             StopButton.Click += StopButton_Click;
         }
@@ -29,6 +30,12 @@ namespace LibVLCSharp.WPF.Sample
             _mediaPlayer = new MediaPlayer(_libVLC);
 
             parent.VideoView.MediaPlayer = _mediaPlayer;
+        }
+
+        private void VideoView_Unloaded(object sender, RoutedEventArgs e)
+        {
+            _mediaPlayer.Dispose();
+            _libVLC.Dispose();
         }
 
         void StopButton_Click(object sender, RoutedEventArgs e)
--- a/samples/LibVLCSharp.WPF.Sample/Example2.xaml.cs
+++ b/samples/LibVLCSharp.WPF.Sample/Example2.xaml.cs
@@ -35,24 +35,26 @@ namespace LibVLCSharp.WPF.Sample
 
         void StopButton_Click(object sender, RoutedEventArgs e)
         {
-            if (VideoView.MediaPlayer.IsPlaying)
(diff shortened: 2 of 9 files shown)


mfkl:
Thanks for this.

I'm wondering if there is a _real_ need for this. Yes, for the HWND case, this is probably a good change but for all the APIs? I'm not sure. 

We don't have bug reports for this, so does it really need fixing?
On the other hand, if we want to go this way, I'd rather throw than fail silently (by returning a default value). This really should almost never happen, and an `ObjectDisposedException` would fit IMO.

It seems other bindings don't do these checks though, so I need to think about it a bit more.

Have you tried to cause a crash in a sample by calling a libvlc api after disposing the object?

And thank you for starting to add the dispose code to the samples, this is indeed something we need to add to all samples to encourage and educate users about this.

jeremyVignelles:
> I'm wondering if there is a _real_ need for this. Yes, for the HWND case, this is probably a good change but for all the APIs? I'm not sure.
> 
> We don't have bug reports for this, so does it really need fixing?

Wondering too, but I think it's a good practice to prevent native crash/unexpected behaviors...

> On the other hand, if we want to go this way, I'd rather throw than fail silently (by returning a default value). This really should almost never happen, and an `ObjectDisposedException` would fit IMO.

Except for that HWND case, where a throw would not be better... Or maybe the fix is to check the NativeReference/IsDisposed from outside the MediaPlayer (in the VideoView)

> It seems other bindings don't do these checks though, so I need to think about it a bit more.

Sure

> Have you tried to cause a crash in a sample by calling a libvlc api after disposing the object?

The modified sample code does crash without the check in the HWND setter.

> And thank you for starting to add the dispose code to the samples, this is indeed something we need to add to all samples to encourage and educate users about this.

👍

mfkl:
Ok, I had a proper look and more time to think about this.

I don't think we should take these changes in at this time, because
- It worsens visibility of the code (expression bodied members are kinda nice),
- This isn't fixing an actual user issue we currently have.

I also looked at similar bindings, namely skiasharp and libgit2sharp, and they don't seem to be handling this case either (it is mentioned though https://github.com/libgit2/libgit2sharp/issues/1445).

_However_, were user reports starting to pop up regarding `AccessViolationException` about this (that is the symptom), I would be completely willing to revisit this, provided we would throw `ObjectDisposedException` instead of early return silently. This error is a signal that user code is incorrect and an exception is the correct way to go IMO.

About the samples dispose code, I will cherry pick your changes as work towards https://code.videolan.org/videolan/LibVLCSharp/-/issues/401.

mvdhorst:
It does solve a part of my issue, when I use multiple mediaplayers, and dispose them, it could occur that the reference become IntPtr.Zero

mfkl:
Could you please open a ticket on gitlab with a repro sample so we can have a look? Thanks.

PR #172, opened 2020-11-06
Wpf eventhandler fixes

SkinnySackboy:
### Description of Change ###

1. The WPF window loaded event can fire multiple times, therefore a check has been added to ensure this is now only executed once as otherwise there is the risk of attempting to show a window more than once (which is illegal in WPF and throws an exception)
2. When the WPF window is closing, event handlers are now unsubscribed

### Issues Resolved ### 
Exceptions thrown by WPF when docking/undocking/showing/hiding a window.

### API Changes ###
None

### Platforms Affected ### 
WPF

No exceptions thrown when docking/undocking window.

Not applicable

### Testing Procedure ###
https://code.videolan.org/videolan/LibVLCSharp/-/issues/384

### PR Checklist ###

### DIFF
--- a/src/LibVLCSharp.WPF/ForegroundWindow.xaml.cs
+++ b/src/LibVLCSharp.WPF/ForegroundWindow.xaml.cs
@@ -102,8 +102,13 @@ namespace LibVLCSharp.WPF
 
         void Wndhost_SizeChanged(object sender, SizeChangedEventArgs e)
         {
-            var locationFromScreen = _bckgnd.PointToScreen(_zeroPoint);
             var source = PresentationSource.FromVisual(_wndhost);
+            if (source == null)
+            {
+                return;
+            }
+
+            var locationFromScreen = _bckgnd.PointToScreen(_zeroPoint);
             var targetPoints = source.CompositionTarget.TransformFromDevice.Transform(locationFromScreen);
             Left = targetPoints.X;
             Top = targetPoints.Y;
--- a/src/LibVLCSharp.WPF/ForegroundWindow.xaml.cs
+++ b/src/LibVLCSharp.WPF/ForegroundWindow.xaml.cs
@@ -125,6 +125,10 @@ namespace LibVLCSharp.WPF
         void Wndhost_Closing(object sender, System.ComponentModel.CancelEventArgs e)
         {
             Close();
+
+            _bckgnd.DataContextChanged -= Background_DataContextChanged;
+            _bckgnd.Loaded -= Background_Loaded;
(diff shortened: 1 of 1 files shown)


src/LibVLCSharp.WPF/ForegroundWindow.xaml.cs
```diff
@@ -58,6 +58,11 @@ void Background_Unloaded(object sender, RoutedEventArgs e)
 
         void Background_Loaded(object sender, RoutedEventArgs e)
         {
+            if (_wndhost != null)
```
mfkl, line 61: Should `_wndhost` be explicitly set to `null` at some point during the `Window` lifecycle? On `Wndhost_Closing` likely.

SkinnySackboy: We can do that, though bear in mind that the `Owner` will still remain populated if `_wndhost` is set to null, unless we choose to null that also.

The reason I used `_wndhost` as a proxy for being initialised (to ensure the logic within the `Background_Loaded` is only executed at most once) was to avoid maintaining additional state inside this class. 

If we decide to set `_wndhost` to 'null' we would also need to add a new bool field `_isInitialized` to ensure 'Background_Loaded' is only called once, because it will no longer be sufficient to check `_wndhost` for being null as a guard.

Let me know if you would like me to make this change.

mfkl: > If we decide to set _wndhost to null

I'm not clear on whether that's actually needed or not, just wondering. None of the fields are `IDisposable` so it may not be necessary. 

What do you think?

SkinnySackboy: Agreed, I don't think it's necessary.

mfkl (approved):

SkinnySackboy:
Is this all ok @mfkl or do you still need something from me?

mfkl:
Thanks @SkinnySackboy!

SkinnySackboy:
Thanks for your help @mfkl

mfkl:
It's live in 3.4.9

SkinnySackboy:
Thanks.

SkinnySackboy:
We have recently upgraded from 3.5.1 to 3.6.1 and unfortunately these issues which had been resolved seem to have been re-introduced.

jeremyVignelles:
@SkinnySackboy : Please open a new issue on our bugtracker so that we don't forget it

SkinnySackboy:
Will do!

PR #175, opened 2020-11-12
net50 SDK : Fixed Nullable reference types-related warnings

jeremyVignelles:
### Description of Change ###
Updated the .net core SDK in use to 50 and fixed build warnings and errors.

Used a newer version of MSBuild.Sdk.Extras

Fixed warnings and errors.

### API Changes ###

 None

### Platforms Affected ### 

- Core (all platforms)

None

Not applicable

### Testing Procedure ###

Try to build

### PR Checklist ###


src/LibVLCSharp/Shared/Helpers/MarshalUtils.cs
```diff
@@ -509,7 +513,7 @@ internal static T PtrToStructure<T>(IntPtr ptr)
 #if NETSTANDARD1_1 || NET40
             return (T)Marshal.PtrToStructure(ptr, typeof(T));
 #else
-            return Marshal.PtrToStructure<T>(ptr);
+            return Marshal.PtrToStructure<T>(ptr)!;
```
jeremyVignelles, line 516: I have a doubt here, but I think that `T` should be written as `T?` on the caller side to allow it to return a null, what do you think ?
Seems like allowing null here would break 20 usage locations.

jeremyVignelles:
Still having many warnings in net50 due to MSBuild.Sdk.Extras, same issue as described in https://github.com/novotnyllc/MSBuildSdkExtras/issues/246 .
It seems that net50 support is not ready yet, but the doc says that MSBuild.Sdk.Extras 3.x should work with net50. It's not available yet though...

src/LibVLCSharp/Shared/Helpers/MarshalUtils.cs
mfkl: `PtrToStructure` would return an empty struct but not null, right?

jeremyVignelles: T might be a ref type, I guess. the compiler was yelling anyway.
Maybe we could add a `struct` constraint

mfkl: > T might be a ref type, I guess.

Could but we don't marshal ref types currently. No strong feelings tho

jeremyVignelles: Then I'd be in favor of adding a struct constraint to the method and remove the assertion here

mfkl: Sure

mfkl:
This should be good to go now. Needs rebase.

mfkl:
macOS logs: https://videolan.visualstudio.com/73ec5719-cee8-4da9-937a-9d8a3993d7f5/_apis/build/builds/3706/logs/20

mfkl:
Cannot begin to repro locally as VS for Mac does not want to use the .NET 5 SDK here.

winkmichael:
> Cannot begin to repro locally as VS for Mac does not want to use the .NET 5 SDK here.

Same here.

mfkl:
Let's get this merged soon.

mfkl:
Or not. There is still an issue on macOS. Guess this will have to wait for .NET 6 or an msbuildsdk extras fix.

mfkl:
Superseded by https://github.com/videolan/libvlcsharp/pull/248. Thanks :)

PR #179, opened 2020-12-14
Fix ci windows

mfkl:


mfkl:
Will open again when fixed.

PR #181, opened 2020-12-21
Avalonia stable

mfkl:
will rebase when https://github.com/videolan/libvlcsharp/pull/180 is merged


mfkl:
Will reopen when 0.10.0 stable is released.

PR #199, opened 2021-02-24
docs: Make libvlc option syntax usage more explicit

mfkl:


jeremyVignelles:
Maybe add something to the How Do I do X? with an example from the doc? I saw there was a new doc website with some parameters examples, we should link it as well

mfkl:
Yeah.. people seem to miss this https://github.com/videolan/libvlcsharp/blob/3.x/docs/libvlc_documentation.md#libvlc-options

jeremyVignelles:
Should the name of the page be renamed from "LibVLC documentation" to something more explicit for new users? If I'm told LibVLCSharp is based on libvlc, I'd probably assume that this doc contains obscure documentation about how it works internally, not something about passing parameters... Or maybe split that in two questions in the How do I do X?

mfkl:
Sure

PR #202, opened 2021-03-27
Update Core.cs

yasinsensoy:
This solves the character problem.

### DIFF
--- a/src/LibVLCSharp/Shared/Core/Core.cs
+++ b/src/LibVLCSharp/Shared/Core/Core.cs
@@ -22,7 +22,7 @@ namespace LibVLCSharp.Shared
                 EntryPoint = "libvlc_get_version")]
             internal static extern IntPtr LibVLCVersion();
 #endif
-            [DllImport(Constants.Kernel32, SetLastError = true)]
+            [DllImport(Constants.Kernel32, SetLastError = true, CharSet = CharSet.Unicode)]
             internal static extern IntPtr LoadLibrary(string dllToLoad);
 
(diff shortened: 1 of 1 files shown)


jeremyVignelles:
Congrats for your first PR 👏 🎉 

Did you try the change? Does it solve your issue?
I think you would also need to rename `LoadLibrary` to `LoadLibraryW` to use the UTF-16 variant of the Win32 API

yasinsensoy:
> Congrats for your first PR 👏 🎉
> 
> Did you try the change? Does it solve your issue?
> I think you would also need to rename to to use the UTF-16 variant of the Win32 `LoadLibrary``LoadLibraryW`API

Thanks. I tried it solved the problem.

jeremyVignelles (approved):
LGTM then

yasinsensoy:
Will this change apply to NuGet?

jeremyVignelles:
Only after it's merged and after we release a new 3.x package

yasinsensoy:
Understood thanks

mfkl:
> When the ExactSpelling field is false, as it is by default in C++ and C#, platform invoke searches for the mangled name first (MessageBoxW), then the unmangled alias (MessageBox) if the mangled name is not found. Notice that Unicode name-matching behavior differs from ANSI name-matching behavior.

https://docs.microsoft.com/en-us/dotnet/framework/interop/specifying-a-character-set

Edited your patch so that `LoadLibraryW` is clearly specified in the code, as it is the one being used.

Merged, thanks.

PR #203, opened 2021-04-05
Feature/output callbacks

mfkl:
### Description of Change ###

Add set output callbacks API bindings to master

### API Changes ###

Added:
 - public bool SetOutputCallbacks(VideoEngine engine, OutputSetup? outputSetup, OutputCleanup? outputCleanup, OutputSetResize? resize,
            UpdateOutput updateOutput, Swap swap, MakeCurrent makeCurrent, GetProcAddress getProcAddress, FrameMetadata? metadata,
            OutputSelectPlane? selectPlane)

### Platforms Affected ### 

- Core (all platforms)


samples/LibVLCSharp.CustomRendering.Direct3D11/Program.cs
```diff
@@ -0,0 +1,682 @@
...
+        static ID3D11Buffer* CreateBuffer(D3D11_BUFFER_DESC bd)
+        {
+            ID3D11Buffer* buffer;
+
+            ThrowIfFailed(_d3dDevice->CreateBuffer(&bd, null, &buffer));
+
+            return buffer;
+        }
+
+        static readonly string shaderStr = @"
```
mfkl, line 444: Should we ship this shader string as part of LibVLCSharp?

samples/LibVLCSharp.CustomRendering.Direct3D11/Program.cs
```diff
@@ -0,0 +1,682 @@
...
+            internal float x;
+            internal float y;
+            internal float z;
+        }
+
+        [StructLayout(LayoutKind.Sequential)]
+        internal struct Texture
+        {
+            internal float x;
+            internal float y;
```
mfkl, line 495: Should we ship these structs as part of LibVLCSharp? They'll likely be needed for all code implementing this API (at least the default use case)

src/LibVLCSharp/Structures/OutputCallbacks.cs
```diff
@@ -0,0 +1,370 @@
...
+    /// <summary>
+    /// Device configuration setup for ouput callbacks
+    /// </summary>
+    [StructLayout(LayoutKind.Sequential)]
+    public readonly struct SetupDeviceConfig
+    {
+        /// <summary>
+        /// set to true if D3D11_CREATE_DEVICE_VIDEO_SUPPORT is needed for D3D11
+        /// </summary>
+        public readonly bool HardwareDecoding;
```
mfkl, line 15: Use properties for struct, better perf

mfkl: done

samples/LibVLCSharp.CustomRendering.Direct3D11/Program.cs
mfkl: Done. opengl will too. This simplifies the sample code.

samples/LibVLCSharp.CustomRendering.Direct3D11/Program.cs
mfkl: Probably not. It's used in the sample app but not in other implementations (like vlc-unity). Let users implement their own structs if needed.

PR #204, opened 2021-04-21
MediaTrackList: Fix crash in indexer binding code

mfkl:
https://code.videolan.org/videolan/vlc-unity/-/issues/104

https://mailman.videolan.org/pipermail/vlc-devel/2021-April/143192.html

### DIFF
--- a/src/LibVLCSharp/Structures/MediaTrack.cs
+++ b/src/LibVLCSharp/Structures/MediaTrack.cs
@@ -284,7 +284,7 @@ namespace LibVLCSharp
         {
             get
             {
-                if (position > Count)
+                if (position >= Count)
                     return null;
 
                 var ptr = Native.LibVLCMediaTrackListAt(NativeReference, (UIntPtr)position);


src/LibVLCSharp/Structures/MediaTrack.cs
```diff
@@ -284,7 +284,8 @@ internal MediaTrackList(IntPtr mediaTrackListPtr) : base(() => mediaTrackListPtr
         {
             get
             {
-                if (position > Count)
+                var count = Count;
+                if (count == 0 || position > count)
```
jeremyVignelles, line 288: Should likely be `if (position >= count)`. What if Count is 1 and `[1]` is called?

mfkl: > What if Count is 1 and `[1]` is called?

Hmm then that'd be an invalid call I guess? Count is 1-based while the indexer (`libvlc_media_tracklist_at`) is 0-based. Making me doubt now :D

mfkl: Added your suggestion just to be safe, though from my understanding that condition would never have ran considering the following list implementation
```
public bool MoveNext()
{
    position++;
    return position < (_mediaTrackList?.Count ?? 0);
}
```

src/LibVLCSharp/Structures/MediaTrack.cs
```diff
@@ -284,7 +284,8 @@ internal MediaTrackList(IntPtr mediaTrackListPtr) : base(() => mediaTrackListPtr
         {
             get
             {
-                if (position > Count)
+                var count = Count;
+                if (count == 0 || position >= count)
```
jeremyVignelles, line 288: Could be simplified to the one-liner :
```suggestion
                if (position >= Count)
```

if `Count == 0` there's no way for an `uint` to be below

mfkl: Sure

src/LibVLCSharp/Structures/MediaTrack.cs
jeremyVignelles: `MoveNext` is called when iterating in a `foreach` loop or when doing linq, but the indexer is not used in that case.

The proper test case would be to call `mediaTrackList[1]` directly, without checking the bounds.

jeremyVignelles (approved):

src/LibVLCSharp/Structures/MediaTrack.cs
mfkl: > MoveNext is called when iterating in a foreach loop or when doing linq, but the indexer is not used in that case.

Yes, it is (but it doesn't matter) :-) The `IEnumerator` states that after `MoveNext` is called when iterating over a list such as the media track list, `Current` is then called, which does `return _mediaTrackList[position]`.

PR #205, opened 2021-04-26
Media: Internally keep a reference to the last set Media

mfkl:
### Description of Change ###

libvlc_media_player_get_media increments the ref count of the media, which may easily
cause a memory leak, as this is very unusual behavior for C# property getter to
have these types of consequences. Users don't have the reflex to dispose every Media returned
by the getter property.

Holding a ref internally in the mediaplayer prevents from incrementing the native ref count
of the media object.

### Issues Resolved ### 

- fixes https://code.videolan.org/videolan/LibVLCSharp/-/issues/442

### API Changes ###
 
 None, but the usage is subtly changed. Docs should probably mentioned it somewhere.

### Platforms Affected ### 

- Core (all platforms)

None

Not applicable

### Testing Procedure ###

I have not tested this thoroughly.

### PR Checklist ###

### DIFF
--- a/src/LibVLCSharp/Shared/MediaPlayer.cs
+++ b/src/LibVLCSharp/Shared/MediaPlayer.cs
@@ -663,11 +663,9 @@ namespace LibVLCSharp.Shared
         /// <returns>true if successful</returns>
         public bool Play()
         {
-            var media = Media;
-            if(media != null)
+            if(_media != null)
             {
-                media.AddOption(Configuration);
-                media.Dispose();
+                _media.AddOption(Configuration);
             }
             return Native.LibVLCMediaPlayerPlay(NativeReference) == 0;
         }
--- a/src/LibVLCSharp/Shared/MediaPlayer.cs
+++ b/src/LibVLCSharp/Shared/MediaPlayer.cs
@@ -631,20 +631,23 @@ namespace LibVLCSharp.Shared
         /// </summary>
         public Media? Media
         {
-            get
-            {
-                return _media;
-            }
+            get => _media;
             set
             {
-                if(_media?.NativeReference != IntPtr.Zero)
+                if(_media != null)
                 {
-                    _media?.Dispose();
+                    _media.Dispose();
                     _media = null;
                 }
 
                 _media = value;
-                Native.LibVLCMediaPlayerSetMedia(NativeReference, value?.NativeReference ?? IntPtr.Zero);
+                
+                if(_media != null)
+                {
+                    _media.Retain();
+                }
+
+                Native.LibVLCMediaPlayerSetMedia(NativeReference, _media?.NativeReference ?? IntPtr.Zero);
             }
         }
 
--- a/src/LibVLCSharp/Shared/MediaPlayer.cs
+++ b/src/LibVLCSharp/Shared/MediaPlayer.cs
@@ -2416,7 +2416,8 @@ namespace LibVLCSharp.Shared
             {
                 if (_gcHandle.IsAllocated)
                     _gcHandle.Free();
-                if (_media?.NativeReference != IntPtr.Zero)
+
+                if (_media != null)
                 {
                     _media?.Dispose();
                     _media = null;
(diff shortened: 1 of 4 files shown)


jeremyVignelles:
This is a breaking change for those who already did that properly. If they Disposed the media by themselves, the media ends up getting disposed multiple times.

Aren't there some cases where the libvlc media player decides to change the media internally? like making it null on Stop or when playing a playlist?

mfkl:
> This is a breaking change for those who already did that properly. If they Disposed the media by themselves, the media ends up getting disposed multiple times.

I know... :-( But trying to find ways to fix it anyway, for 3.x. If it's not possible, the fix will be 4.x only.

> Aren't there some cases where the libvlc media player decides to change the media internally? like making it null on Stop or when playing a playlist?

Tried to find these cases but failed. We don't expose the medialistplayer object so that helps somewhat to reduce potentially affected use cases.

src/LibVLCSharp/Shared/MediaPlayer.cs
```diff
@@ -635,10 +633,19 @@ public MediaPlayer(Media media)
         {
             get
             {
-                var mediaPtr = Native.LibVLCMediaPlayerGetMedia(NativeReference);
-                return mediaPtr == IntPtr.Zero ? null : new Media(mediaPtr);
+                return _media;
+            }
+            set
+            {
+                if(_media?.NativeReference != IntPtr.Zero)
```
jeremyVignelles, line 640: Either media is null or it's not, but if it's not null, I don't think its native reference can be Zero. Either way, it's up to _media.Dispose() to check that (same in MediaPlayer.Dispose)

mfkl:
Yes.. I see what you mean :-) My problem with this design is that it creates a layer of abstraction over LibVLC and I'd like to avoid that because it is bound to grow with time, and become problematic as the underlying LibVLC API changes.

This awkward libvlcsharp API won't be fixed for 3.x, because as you noted it's a subtle breaking change which has a high risk of breaking user-code. However, we can definitly fixed it for 4.x and document it as a breaking change.

I don't understand yet why LibVLC must increment the ref count when calling get_media on a media that already has a refcount > 1. Maybe changing LibVLC is an option? 

`libvlc_media_player_new_from_media` also increments the media ref count, as well as `libvlc_media_player_set_media`...

jeremyVignelles:
I don't think so. If you don't count the usages, libvlc won't be able to know that the media is still kept for later use and that it is not expected to be destroyed

mfkl:
So I tried to call media_release repeatedly and it seems that libvlc is fine with it... no crash.

2 solutions that I have so far. Aiming for something _simple_...
- what's currently in this branch. Proxy the media by keeping a single ref in the binding to prevent from incrementing the native ref count.
**Advantage**: Can use that property getter (which is _really_ convenient) as much as we want, like a normal C# property.
**Disadvantage**: ???
- rename the property from `mediaplayer.Media` to `mediaplayer.NewMedia` (or make it a method), and document that users need to dispose that.
**Advantage**: Doesn't circumvent the native API, uses the native ref count as intended.
**Disadvantage**: People won't read the doc and dispose it, so it will cause leak. That's for sure.

Still looking for other solutions. No special hurry as this is v4 anyway.

jeremyVignelles:
1. Disadvantage : You can't reuse Media for A/B loops
2. I'd be in favor of making it a method. Renaming the property doesn't really add value.
3. We could add IDisposable annotations (see my proposal [here](https://github.com/dotnet/runtime/issues/29631) ) and recommend using an analyzer like [IDisposableAnalzyer](https://github.com/DotNetAnalyzers/IDisposableAnalyzers) that would spot such issues, but [work is needed here](https://github.com/DotNetAnalyzers/IDisposableAnalyzers/issues/126)

mfkl:
Added a failing test.

While this approach may seem like a good idea, for now there is a still a few undiscovered gotchas. And the retain/dispose routine is different if you create a mediaplayer from a media or if you set a media on a mediaplayer instance...

mfkl:
There needs to be a solution for this problem for libvlcsharp 4.

mfkl:
Gotta merge this one soon.

mfkl:
Introduces too much magic. Closing for now.

PR #209, opened 2021-05-04
core: remove dispose check already handled in base class

mfkl:


src/LibVLCSharp/Shared/RendererDiscoverer.cs
```diff
@@ -191,9 +188,6 @@ public class RendererItem : Internal
         /// <param name="disposing">true if called from a method</param>
         protected override void Dispose(bool disposing)
         {
-            if (IsDisposed || NativeReference == IntPtr.Zero)
```
jeremyVignelles, line 194: If you're just calling `base.Dispose(disposing)`,  you could jut remove the override altogether

mfkl: Done

jeremyVignelles (approved):

PR #211, opened 2021-05-31
Core: Fix resolving libvlc location for single file self contained apps

mfkl:
### Description of Change ###

`Assembly.Location` is deprecated and returns `string.Empty` in .NET 5 and later when using Single file self contained apps. 

See https://docs.microsoft.com/en-us/dotnet/core/deploying/single-file#api-incompatibility

### Issues Resolved ### 

- fix https://code.videolan.org/videolan/LibVLCSharp/-/issues/470

### API Changes ###
 
 None

### Platforms Affected ### 

- Core (all platforms)

Haven't tried Linux/macOS

None

Not applicable

### Testing Procedure ###

Checked that this fixes running a self contained single app net5 app sample on Windows.

### PR Checklist ###

### DIFF
--- a/src/LibVLCSharp/Shared/Core/Core.cs
+++ b/src/LibVLCSharp/Shared/Core/Core.cs
@@ -72,7 +72,15 @@ namespace LibVLCSharp.Shared
                 arch = PlatformHelper.IsX64BitProcess ? ArchitectureNames.Win64 : ArchitectureNames.Win86;
             }
 
-            var libvlcDirPath1 = Path.Combine(Path.GetDirectoryName(typeof(LibVLC).Assembly.Location),
+            var libvlcAssemblyLocation = typeof(LibVLC).Assembly.Location;
+
+#if !NET40
+            if (string.IsNullOrEmpty(libvlcAssemblyLocation)) /* .NET 5 (single file / self contained) and later */
+            {
+                libvlcAssemblyLocation = AppContext.BaseDirectory;
+            }
+#endif
+            var libvlcDirPath1 = Path.Combine(Path.GetDirectoryName(libvlcAssemblyLocation),
                 Constants.LibrariesRepositoryFolderName, arch);
 
             var libvlccorePath1 = LibVLCCorePath(libvlcDirPath1);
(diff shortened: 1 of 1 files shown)


src/LibVLCSharp/Shared/Core/Core.cs
```diff
@@ -72,7 +72,16 @@ static List<(string libvlccore, string libvlc)> ComputeLibVLCSearchPaths()
                 arch = PlatformHelper.IsX64BitProcess ? ArchitectureNames.Win64 : ArchitectureNames.Win86;
             }
 
-            var libvlcDirPath1 = Path.Combine(Path.GetDirectoryName(typeof(LibVLC).Assembly.Location),
+            var libvlcAssemblyLocation = typeof(LibVLC).Assembly.Location;
+
+            if (string.IsNullOrEmpty(libvlcAssemblyLocation)) /* .NET 5 and later */
+            {
+#if !NET40
```
jeremyVignelles, line 79: Why not around the if ?
Location should not be an empty string in NET40, but what if it actually does ? Shouldn't we check the returned values ?
For example, Path.GetDirectoryName may return null or an empty string : https://docs.microsoft.com/en-us/dotnet/api/system.io.path.getdirectoryname?view=net-5.0#System_IO_Path_GetDirectoryName_System_String_

mfkl: > Why not around the if ?

Doesn't really make a difference.

> Location should not be an empty string in NET40, but what if it actually does ? Shouldn't we check the returned values ?

We could, but while `Assembly.Location` _could_ be an empty string on an edge case
>  If the assembly is loaded from a byte array, such as when using Load(Byte[]), the value returned is an empty string ("")

`AppContext.BaseDirectory` being null or empty would mean the dotnet runtime can't function, seeing how pervasive its used in there.

> For example, Path.GetDirectoryName may return null or an empty string

> Directory information for path, or null if path denotes a root directory or is null. Returns Empty if path does not contain directory information.

Checking for this seems a bit too defensive for me. I'd say we wait until (if?) this problem occurs before we start checking for it?

jeremyVignelles: 🆗

PR #225, opened 2021-07-26
MediaElement XF: Improve track management - Fixes #196

egbakou:
### Description of Change ###

An Overlay View is used to display all tracks like it is done on VLC Android. User can select any track from the tracks list which will be automatically applied to the media.
This feature was initially implemented using Popup.

### Issues Resolved ### 

- fixes https://code.videolan.org/videolan/LibVLCSharp/-/issues/196

### API Changes ###

Added:
 - VideoTracksManager 

Changed:
 - MediaPlayerElement.xaml.cs
 - PlaybackControls.xaml.cs
 - Strings.Designer.cs 
 - Generic.xaml 
 - MediaPlayerElementManager.cs
 
 Removed:
 - SelectTrackAsync method of PlaybackControls.xaml.cs
 

### Platforms Affected ### 

- iOS
- Android

Before:
Android          |  iOS
:-------------------------:|:-------------------------:
 |  
 |  

After:
Android          |  iOS
:-------------------------:|:-------------------------:
 |  
 |  

### Testing Procedure ###
None

### PR Checklist ###


mfkl:
Can you rebase your branch on 3.x latest please?
Have you tested the Xamarin.Forms GTK, macOS and WPF backends?

mfkl:
Can you please update https://code.videolan.org/videolan/LibVLCSharp/-/blob/3.x/docs/media_player_element.md?

mfkl:
image

I think the background of the titles (video tracks, audio tracks, subtitles) should also be darker, to match the list item. This would match the current vlc android app design. What do you think?

src/LibVLCSharp.Forms/Shared/Themes/Generic.xaml
```diff
@@ -148,6 +148,24 @@
         </Setter>
     </Style>
 
+    <Style x:Key="VideoTracksSelectionButtonStyle" BasedOn="{StaticResource ButtonStyle}" TargetType="Button">
+        <Setter Property="IsVisible" Value="{TemplateBinding IsVideoTracksSelectionButtonVisible}" />
+        <Setter Property="FontFamily" Value="{StaticResource FontAwesomeRegular}" />
+        <Setter Property="Text" Value="{x:Static fontawesome:FontAwesomeIcons.FileVideo}" />
```
mfkl, line 154: I'm not sure I'm seeing that button anywhere?

mfkl:
When you double click (android simulator), row turns orange. Is that expected?
image

mfkl:
I'm thinking a white/orange checkmark ✔️ could be more visible than the (*) for selected tracks, what do you think?

mfkl:
Thinking the `List` or `ListAlt` icon might be more suited than the file one. Looking at https://fontawesome.com/v5/cheatsheet if you have better ideas :-)

src/LibVLCSharp.Forms/Shared/Themes/Generic.xaml
egbakou: Yeah, I will remove it.

egbakou:
> When you double click (android simulator), row turns orange. Is that expected?
> image



> When you double click (android simulator), row turns orange. Is that expected?
> image
This is the default behavior of Listviews when the item is selected with a long press or a double-click.
To change the color, we need to implement custom ViewCell.
I don't know if it is important. What do you think ?

egbakou:
> image
> 
> I think the background of the titles (video tracks, audio tracks, subtitles) should also be darker, to match the list item. This would match the current vlc android app design. What do you think?

I agree with you.

egbakou:
> I'm thinking a white/orange checkmark ✔️ could be more visible than the (*) for selected tracks, what do you think?

I have to try if we can set the checkmark from code behind.

egbakou:
> Can you rebase your branch on 3.x latest please?
Okay
> Have you tested the Xamarin.Forms GTK, macOS and WPF backends?
No, I haven't. Should I do it ?

mfkl:
> > When you double click (android simulator), row turns orange. Is that expected?
> > image
> 
> This is the default behavior of Listviews when item is selected using a long press or a double-click.
> To change the color, we need to implement custom ViewCell.
> I don't know if it is important. What do you think ?

I guess you can leave it as is if the user can customize it themselves.

mfkl:
> > Have you tested the Xamarin.Forms GTK, macOS and WPF backends?
> 
> No, I haven't. Should I do it ?

You could, I'd be curious to see how that looks like. Technically we support those platforms for the MediaElement.

egbakou:
> 
> 
> > image
> > I think the background of the titles (video tracks, audio tracks, subtitles) should also be darker, to match the list item. This would match the current vlc android app design. What do you think?
> 
> I agree with you. It's done.

DarkBg-Tracks-Android

egbakou:
> 
> 
> Can you please update https://code.videolan.org/videolan/LibVLCSharp/-/blob/3.x/docs/media_player_element.md?

Done

src/LibVLCSharp.Forms/Shared/PlaybackControls.xaml.cs
```diff
@@ -821,14 +859,126 @@ private void OnAudioTracksChanged(object sender, EventArgs e)
...
+            var tracks = manager.Tracks;
+            var currentTrackId = manager.CurrentTrackId;
+            if (tracks == null || track.Name.Contains("*"))
+                return;
+            var found = false;
+            var index = 0;
+            foreach (var trackDescription in tracks)
+            {
+                index += 1;
+                if (GetTrackName(trackDescription, currentTrackId) == track.Name)
```
jeremyVignelles, line 923: I don't get why you compare on the name, where `*` is appended if it is the current track... Why not using the Track Id ?

src/LibVLCSharp.Forms/Shared/PlaybackControls.xaml.cs
```diff
@@ -821,14 +859,126 @@ private void OnAudioTracksChanged(object sender, EventArgs e)
...
+
+        private void SelectTrack(TracksManager manager, TrackDescription track, ListView? tracksListview)
+        {
+            var tracks = manager.Tracks;
+            var currentTrackId = manager.CurrentTrackId;
+            if (tracks == null || track.Name.Contains("*"))
+                return;
+            var found = false;
+            var index = 0;
+            foreach (var trackDescription in tracks)
```
jeremyVignelles, line 920: Seems like this foreach could have been replaced by ` var foundTrack = tracks.FirstOrDefault(t => ...)` and a if/else depending on if it was found.

src/LibVLCSharp.Forms/Shared/PlaybackControls.xaml.cs
```diff
@@ -821,14 +859,126 @@ private void OnAudioTracksChanged(object sender, EventArgs e)
...
+        private void UpdateTracksListviewItemsSource(TrackDescription selectedTrack, ListView? trackListView)
+        {
+            if (trackListView != null)
+            {
+                var itemSources = (ObservableCollection<TrackDescription>)trackListView.ItemsSource;
+                var previousTrack = itemSources.FirstOrDefault(t => t.Name[t.Name.Length - 1] == '*');
+
+                // Put "*" at the end of the selected track
+                itemSources.RemoveAt(selectedTrack.Id);
+                itemSources.Insert(selectedTrack.Id, new TrackDescription(selectedTrack.Id, $"{selectedTrack.Name} *"));
```
jeremyVignelles, line 946: I'm not a huge fan of replacing the item in the collection for the purpose of adding an asterisk for the selected track.

The asterisk should have been a converter or something, to allow different styles to show the current track the way they want.

mfkl: see https://github.com/videolan/libvlcsharp/pull/225#issuecomment-887233659

src/LibVLCSharp.Forms/Shared/PlaybackControls.xaml.cs
```diff
@@ -98,20 +104,26 @@ public PlaybackControls()
...
 
         private MediaPlayerElementManager Manager { get; }
-        private Button? AudioTracksSelectionButton { get; set; }
+        private Button? TracksButton { get; set; }
         private Button? CastButton { get; set; }
-        private Button? ClosedCaptionsSelectionButton { get; set; }
         private VisualElement? ControlsPanel { get; set; }
         private VisualElement? ButtonBar { get; set; }
         private VisualElement? UnLockControlsPanel { get; set; }
+        private VisualElement? TracksOverLayView { get; set; }
```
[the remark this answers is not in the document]
egbakou: Done locally

src/LibVLCSharp.Forms/Shared/PlaybackControls.xaml.cs
```diff
@@ -821,14 +859,126 @@ private void OnAudioTracksChanged(object sender, EventArgs e)
...
+        {
+            var tracks = manager.Tracks;
+            var currentTrackId = manager.CurrentTrackId;
+            if (tracks == null || track.Name.Contains("*"))
+                return;
+            var found = false;
+            var index = 0;
+            foreach (var trackDescription in tracks)
+            {
+                index += 1;
```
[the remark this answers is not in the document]
egbakou: Yeah, sorry !

src/LibVLCSharp.Forms/Shared/PlaybackControls.xaml.cs
egbakou: The trackDescription Ids are not are not ordered like 0,1,2  so that I can index my Itemsource.
For VideoTrack IDs for example, you can see 0, -1, 3.
So when creating the Itemsource for a ListView, I change the index: -1 become 0, -1 become 1, ...
That is why I make the comparison using this name.
You will notice that currently in the 3.x branch, the Track Id is used for the comparaion and it's wrong.

src/LibVLCSharp.Forms/Shared/PlaybackControls.xaml.cs
egbakou: PR updated.

src/LibVLCSharp.Forms/Shared/PlaybackControls.xaml.cs
```diff
@@ -590,6 +588,21 @@ public bool IsAudioTracksSelectionButtonVisible
             set => SetValue(IsAudioTracksSelectionButtonVisibleProperty, value);
         }
 
+        /// <summary>
+        /// Identifies the <see cref="IsVideoTracksSelectionButtonVisible"/> dependency property.
+        /// </summary>
+        public static readonly BindableProperty IsVideoTracksSelectionButtonVisibleProperty = BindableProperty.Create(
```
mfkl, line 594: Is this code still relevant? Ditto for `IsAudioTracksSelectionButtonVisibleProperty` and `IsAudioTracksSelectionButtonVisible`

src/LibVLCSharp.Forms/Shared/PlaybackControls.xaml.cs
```diff
@@ -590,6 +588,21 @@ public bool IsAudioTracksSelectionButtonVisible
...
+            nameof(IsVideoTracksSelectionButtonVisible), typeof(bool), typeof(PlaybackControls), true,
+            propertyChanged: IsVideoTracksSelectionButtonVisiblePropertyChanged);
+        /// <summary>
+        /// Gets or sets a value indicating whether the audio tracks selection button is shown.
```
mfkl, line 598: incorrect copy pasted comment

mfkl:
It's good you added a new image with the updated UI, but it seems heavy compared to the other assets. Can you please either reduce its size and/or minify it? Thank you.

mfkl:
Breaking changes:
- Adding an absolute layout on top may have unknown consequences for users who have customized the control,
- deleted AudioTracksSelectionButtonStyle,
- deleted ClosedCaptionsSelectionButtonStyle.

Our [versioning strategy](https://github.com/videolan/libvlcsharp/blob/3.x/docs/versioning.md#breaking-changes) accounts for that, however that means we only get 3 more breaking changes minor version bumps. After that, since we won't bump to 4.0 until libvlc 4 is out, we won't be able to stick to our versioning strategy :-/

For 4.0 we may want to revisit this strategy to allow more flexibility on our side /cc @jeremyVignelles 

This PR won't apply nicely to the master branch as the media tracks API has been modified for v4.

jeremyVignelles:
> Our [versioning strategy](https://github.com/videolan/libvlcsharp/blob/3.x/docs/versioning.md#breaking-changes) accounts for that, however that means we only get 3 more breaking changes minor version bumps. After that, since we won't bump to 4.0 until libvlc 4 is out, we won't be able to stick to our versioning strategy :-/

I know that it's not intuitive, but having a 3.10 is perfectly fine.

> For 4.0 and onwards we may want to revisit this strategy to allow more flexibility on our side /cc @jeremyVignelles 

Once explained, the versionning strategy we chose make sense, but is not intuitive given the tools we have : Users of 3.x will be prompted to switch to 4.0 when it will be out...

Because we don't know what breaking change will be done in libvlc 5.x, I'm not sure we can safely have a LVS version that supports multiple libvlc API versions...

egbakou:
> It's good you added a new image with the updated UI, but it seems heavy compared to the other assets. Can you please either reduce its dimensions and/or minify it? Thank you.

Android          |  iOS
:-------------------------:|:-------------------------:
 |

src/LibVLCSharp.Forms/Shared/PlaybackControls.xaml.cs
egbakou: Removed.

src/LibVLCSharp.Forms/Shared/PlaybackControls.xaml.cs
egbakou: Updated

mfkl:
> > It's good you added a new image with the updated UI, but it seems heavy compared to the other assets. Can you please either reduce its dimensions and/or minify it? Thank you.
> 
> Android	iOS
> 	

I meant the `src/assets/media-element.png` file that is part of this PR :-)

egbakou:
> but it seems heavy compared to the other assets

Can I change the extension to jpg ?

egbakou:
> > > It's good you added a new image with the updated UI, but it seems heavy compared to the other assets. Can you please either reduce its dimensions and/or minify it? Thank you.
> > 
> > 
> > Android	iOS
> > 	
> 
> I meant the `src/assets/media-element.png` file that is part of this PR :-)

Its size is now 265 kb :-)

src/LibVLCSharp.Forms/Shared/PlaybackControls.xaml.cs
```diff
@@ -821,14 +813,113 @@ private void OnAudioTracksChanged(object sender, EventArgs e)
...
+
+        private void UpdateTracksListviewItemsSource(TrackViewModel selectedTrack, ListView? trackListView)
+        {
+            if (trackListView != null)
+            {
+                var itemSources = (ObservableCollection<TrackViewModel>)trackListView.ItemsSource;
+                var previousTrack = itemSources.FirstOrDefault(t => t.Selected);
+
+                itemSources.ElementAt(previousTrack.Id).Selected = false;
+                itemSources.ElementAt(selectedTrack.Id).Selected = true;
```
jeremyVignelles, line 893:
```suggestion
                previousTrack.Selected = false;
                selectedTrack.Selected = true;
```

You already have the viewModel object references, you don't need to perform another lookup in the object.

src/LibVLCSharp.Forms/Shared/PlaybackControls.xaml.cs
```diff
@@ -821,14 +813,113 @@ private void OnAudioTracksChanged(object sender, EventArgs e)
...
+                manager.CurrentTrackId = -1;
+            }
+        }
+
+        private void UpdateTracksListviewItemsSource(TrackViewModel selectedTrack, ListView? trackListView)
+        {
+            if (trackListView != null)
+            {
+                var itemSources = (ObservableCollection<TrackViewModel>)trackListView.ItemsSource;
+                var previousTrack = itemSources.FirstOrDefault(t => t.Selected);
```
jeremyVignelles, line 890: Either you use `.First()` here, or you need to handle the (unlikely) case where previousTrack is null

src/LibVLCSharp.Forms/Shared/Themes/Generic.xaml
```diff
@@ -239,71 +206,151 @@
...
+                            <Grid>
+                                <StackLayout Grid.Row="0" HorizontalOptions="FillAndExpand">
+                                    <Label Style="{StaticResource TrackTypeLabelStyle}" x:Name="AudioTracksLabel" />
+                                    <ListView x:Name="AudioTracksListView" ItemTemplate="{StaticResource TrackViewCellDataTemplate}" Style="{StaticResource TracksListViewStyle}" />
+                                </StackLayout>
+                                <StackLayout Grid.Row="1" HorizontalOptions="FillAndExpand">
+                                    <Label Style="{StaticResource TrackTypeLabelStyle}" x:Name="VideoTracksLabel" />
+                                    <ListView x:Name="VideoTracksListView" ItemTemplate="{StaticResource TrackViewCellDataTemplate}" Style="{StaticResource TracksListViewStyle}" />
+                                </StackLayout>
+                                <StackLayout Grid.Row="2" HorizontalOptions="FillAndExpand">
```
mfkl, line 346: Wondering if we shouldn't hide these if there are 0 subtitle tracks (or 0 audio/video tracks, unlikely for video at least).
But then one is dynamically added, it should appear again. What do you think?

The benefit might not be worth the added complexity though.

src/LibVLCSharp.Forms/Shared/PlaybackControls.xaml.cs
```diff
@@ -821,14 +813,113 @@ private void OnAudioTracksChanged(object sender, EventArgs e)
...
+        private ObservableCollection<TrackViewModel> LoadTracks(TracksManager manager)
         {
-            await SelectTrackAsync(Manager.Get<AudioTracksManager>(), ResourceManager.GetString(nameof(Strings.AudioTracks)));
+            var allTracks = new ObservableCollection<TrackViewModel>();
+            var tracks = manager.Tracks;
+            try
+            {
+                var currentTrackId = manager.CurrentTrackId;
+                var index = -1;
+                if (tracks != null)
```
jeremyVignelles, line 905: This should be tested right after the `var tracks = manager.Tracks;` line (return immediately in case it's null.

src/LibVLCSharp.Forms/Shared/TrackViewModel.cs
```diff
@@ -0,0 +1,55 @@
...
+{
+    /// <summary>
+    /// Defines the <see cref="TrackViewModel" />.
+    /// </summary>
+    public class TrackViewModel : INotifyPropertyChanged
+    {
+        /// <summary>
+        /// Track description Id
+        /// </summary>
+        public int Id { get; set; }
```
jeremyVignelles, line 14: It's not an id, it's an index, not to be confused with the track id in the VLC side.
Can you explain again why the VLC track id doesn't fit our need?

jeremyVignelles: If we can't use the VLC track id, then we should:
- Change the name here to avoid confusion
- Document in the comments why it's not the same number

src/LibVLCSharp.Forms/Shared/Themes/Generic.xaml
jeremyVignelles: > if one is dynamically added, it should appear again
Doesn't seem complex if it's a data binding with a Converter

Can tracks be added dynamically ? If so, can they be removed ? If it can be removed, what's the stability of an index-based track identification?

mfkl: > Can tracks be added dynamically ?  If so, can they be removed ? 

The user could add an external .srt file for example. I think some streams can dynamically add/remove ES too...

>  If it can be removed, what's the stability of an index-based track identification?

That's fixed in v4 (one of the reasons this PR won't apply nicely to master)
https://github.com/videolan/libvlcsharp/blob/a6816208f29792685373eb556034721ac0684681/src/LibVLCSharp/Structures/MediaTrack.cs#L23

src/LibVLCSharp.Forms/Shared/TrackViewModel.cs
```diff
@@ -0,0 +1,55 @@
...
+                NotifyPropertyChanged();
+            }
+        }
+
+        /// <summary>
+        /// TrackViewModel constructor
+        /// </summary>
+        /// <param name="id">Track description Id</param>
+        /// <param name="name">Track description</param>
+        public TrackViewModel(int id, string name)
```
mfkl, line 40: this should be `internal`

src/LibVLCSharp.Forms/Shared/PlaybackControls.xaml.cs
```diff
@@ -821,14 +813,113 @@ private void OnAudioTracksChanged(object sender, EventArgs e)
...
+        }
+
+        private void SelectTrack(TracksManager manager, TrackViewModel track, ListView? tracksListview)
+        {
+            var tracks = manager.Tracks;
+            var currentTrackId = manager.CurrentTrackId;
+            if (tracks == null || track.Selected)
+                return;
+
+            var foundTrack = tracks.FirstOrDefault(t => t.Name == track.Name);
```
jeremyVignelles, line 873: Still a string comparison here. Can we be sure that the name is unique?

src/LibVLCSharp.Forms/Shared/PlaybackControls.xaml.cs
```diff
@@ -821,14 +813,113 @@ private void OnAudioTracksChanged(object sender, EventArgs e)
...
+            if (foundTrack.Name != null)
```
jeremyVignelles, line 874: If a match is not found, foundTrack is null, not its Name property.

src/LibVLCSharp.Forms/Shared/PlaybackControls.xaml.cs
egbakou: I don't really know is the names come from the Media will always be unique. May be in my ViewModel, I must include the TrackId in addition to the custom Id ?

src/LibVLCSharp.Forms/Shared/PlaybackControls.xaml.cs
egbakou: foundTrack is a struct. Struct are considered value types, and by definition can't be null.

jeremyVignelles: Please disregard my comment then.
How do you expect the name to be null then? Do you display a "Disable" button that is bound to a fake ViewModel?

egbakou: If a match is not found, all string properties will be null. That's why I do the null check on the name.
I think These checks should be removed since the deleted the "*"  in front of the track name (The previous implementation of the Track feature).
We have to replace FirstOrDefault() by First() also because I don't see why foundTrack can not exist.

src/LibVLCSharp.Forms/Shared/PlaybackControls.xaml.cs
egbakou: If Selected is not updated here, It will always remain false because It is used to hide or display the check buttons on the ListView.

jeremyVignelles: I'm not telling you not to set the Selected property, I'm telling you that you don't need to search for things in the itemsSource (espacially based on the index), because you already have the ViewModel instance

egbakou: Ooh sorry. You're right. I just removed it.

src/LibVLCSharp.Forms/Shared/Themes/Generic.xaml
egbakou: > Wondering if we shouldn't hide these if there are 0 subtitle tracks (or 0 audio/video tracks, unlikely for video at least).
> But then, if one is dynamically added, it should appear again. What do you think?

Good idea ! I am currently updating the PR :-)

egbakou:
I updated the PR  :
- The ViewModel's Id and the Track Id are now the same
- ListViews are hidden when items source is empty
- Unnecessary checks removed
- Track Id is used for comparison
- Use of First() instead of FirstOrDefault()

mfkl:
I somehow managed to have this line throw `var foundTrack = tracks.First(t => t.Id == track.Id);` by clicking furiously on the UI. I guess the event through just before the media finished playing, which is why `manager.Tracks` reported only `-1, disable` as a track option (but I managed to select one just before they were cleared by the playback finishing).

Probably not a real issue... just FYI

mfkl (approved):

egbakou:
> I somehow managed to have this line throw `var foundTrack = tracks.First(t => t.Id == track.Id);` an `InvalidOperationException` by clicking furiously on the UI. I guess the event went through just before the media finished playing, which is why `manager.Tracks` reported only `-1, disable` as a track option (but I managed to select one just before they were cleared by the playback finishing).
> 
> Probably not a real issue... just FYI

Thanks for the information. I will handle the Exception by a try clause.

egbakou:
> Thank you @egbakou! :-)

You're welcome :-)
Thank you @mfkl , as well as you @jeremyVignelles.

PR #227, opened 2021-08-08
WPF: Fix VideoView Content (ForegroundWindow) losing winHost hook and…

Yelo420:
… disappearing (Issue #491)

### Description of Change ###

The problem is that the ForegroundWindow is not prepared for an Unload -> Load triggered by the Tab control.
In the Wndhost_LayoutUpdated event an error is thrown, triggered by this line:
var locationFromScreen = _bckgnd.PointToScreen(_zeroPoint);
https://code.videolan.org/videolan/LibVLCSharp/-/blob/3.x/src/LibVLCSharp.WPF/ForegroundWindow.cs#L112
The problem here is that the Wndhost_LayoutUpdated is thrown just before the Background_Unloaded event, even though our anchor "_bckgnd" has already been unloaded at this point (i.e. is no longer connected to a presentation source).
So we can avoid this error by simply asking for this in advance. Like this:
if (PresentationSource.FromVisual(_bckgnd) != null)
Additionally, the ForegroundWindow is now unattached from our anchor and hidden due to the Background_Unloaded event.
So we extend the first condition in the Background_Loaded event to ask if the ForegroundWindow is visible https://code.videolan.org/videolan/LibVLCSharp/-/blob/3.x/src/LibVLCSharp.WPF/ForegroundWindow.cs#L72
With this we make sure that the ForegroundWindow is attached to our anchor again.

### Issues Resolved ### 

- fixes #491

### API Changes ###

 
 None

### Platforms Affected ### 

- Windows WPF

None

Not applicable

### Testing Procedure ###
Manual test

### PR Checklist ###

- [ x] Rebased on top of the target branch at time of PR
- [ x] Changes adhere to coding standard

### DIFF
--- a/src/LibVLCSharp.WPF/ForegroundWindow.cs
+++ b/src/LibVLCSharp.WPF/ForegroundWindow.cs
@@ -69,7 +69,7 @@ namespace LibVLCSharp.WPF
 
         void Background_Loaded(object sender, RoutedEventArgs e)
         {
-            if (_wndhost != null)
+            if (_wndhost != null && IsVisible)
             {
                 return;
             }
@@ -109,11 +109,14 @@ namespace LibVLCSharp.WPF
 
         void Wndhost_LayoutUpdated(object? sender, EventArgs e)
         {
-            var locationFromScreen = _bckgnd.PointToScreen(_zeroPoint);
-            var source = PresentationSource.FromVisual(_wndhost);
-            var targetPoints = source.CompositionTarget.TransformFromDevice.Transform(locationFromScreen);
-            Left = targetPoints.X;
-            Top = targetPoints.Y;
+            if (PresentationSource.FromVisual(_bckgnd) != null)
+            {
+                var locationFromScreen = _bckgnd.PointToScreen(_zeroPoint);
+                var source = PresentationSource.FromVisual(_wndhost);
+                var targetPoints = source.CompositionTarget.TransformFromDevice.Transform(locationFromScreen);
+                Left = targetPoints.X;
+                Top = targetPoints.Y;
+            }
         }
 
         void Wndhost_SizeChanged(object sender, SizeChangedEventArgs e)


src/LibVLCSharp.WPF/ForegroundWindow.cs
```diff
@@ -69,7 +69,7 @@ void Background_Unloaded(object sender, RoutedEventArgs e)
 
         void Background_Loaded(object sender, RoutedEventArgs e)
         {
-            if (_wndhost != null)
+            if (_wndhost != null && IsVisible)
```
mfkl, line 72: Do we want `IsEnabled` here instead?

src/LibVLCSharp.WPF/ForegroundWindow.cs
```diff
@@ -109,11 +109,14 @@ void Background_Loaded(object sender, RoutedEventArgs e)
 
         void Wndhost_LayoutUpdated(object? sender, EventArgs e)
         {
-            var locationFromScreen = _bckgnd.PointToScreen(_zeroPoint);
-            var source = PresentationSource.FromVisual(_wndhost);
-            var targetPoints = source.CompositionTarget.TransformFromDevice.Transform(locationFromScreen);
-            Left = targetPoints.X;
-            Top = targetPoints.Y;
+            if (PresentationSource.FromVisual(_bckgnd) != null)
```
mfkl, line 112: To match the style just below, I'd do this.

```suggestion
            if (PresentationSource.FromVisual(_bckgnd) == null)
                return;
```

src/LibVLCSharp.WPF/ForegroundWindow.cs
jeremyVignelles: Not sure what you mean here? if it's hidden by the tab control, I'm not sure IsEnabled will be false

mfkl:
Thank you for your contribution. I just merged your changes here https://github.com/videolan/libvlcsharp/commit/cd9a51bc18218cf6b327ace8429dcede1d9ba3ba. This should be available shortly as a nightly build and we will make a stable release soon too.

PR #229, opened 2021-08-16
MediaElement XF: Fixes customization regression

egbakou:
### Description of Change ###

Add IsLockButtonVisible and IsTracksButtonVisible properties which allow to hide or show the LockButton and the TracksButton.
### Issues Resolved ### 

- fixes https://code.videolan.org/videolan/LibVLCSharp/-/issues/494

### API Changes ###

 
 None

### Platforms Affected ### 

- iOS
- Android

None

image

### PR Checklist ###

### DIFF
--- a/src/LibVLCSharp.Forms/Shared/Themes/Generic.xaml
+++ b/src/LibVLCSharp.Forms/Shared/Themes/Generic.xaml
@@ -182,6 +182,7 @@
 
     <Style x:Key="LockButtonStyle" BasedOn="{StaticResource ButtonStyle}" TargetType="Button">
         <Setter Property="Text" Value="{x:Static fontawesome:FontAwesomeIcons.UnlockAlt}" />
+        <Setter Property="IsVisible" Value="{TemplateBinding IsLockButtonVisible}" />
     </Style>
 
     <Style x:Key="UnLockButtonStyle" BasedOn="{StaticResource ButtonStyle}" TargetType="Button">
@@ -190,6 +191,7 @@
 
     <Style x:Key="TracksButtonStyle" BasedOn="{StaticResource ButtonStyle}" TargetType="Button">
         <Setter Property="Text" Value="{x:Static fontawesome:FontAwesomeIcons.ListAlt}" />
+        <Setter Property="IsVisible" Value="{TemplateBinding IsTracksButtonVisible}" />
     </Style>
 
     <Style x:Key="AspectRatioButtonStyle" BasedOn="{StaticResource ButtonStyle}" TargetType="Button">
--- a/src/LibVLCSharp.Forms/Shared/PlaybackControls.xaml.cs
+++ b/src/LibVLCSharp.Forms/Shared/PlaybackControls.xaml.cs
@@ -574,6 +574,34 @@ namespace LibVLCSharp.Forms.Shared
         }
 
         /// <summary>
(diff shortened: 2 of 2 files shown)


mfkl (approved):

src/LibVLCSharp.Forms/Shared/Themes/Generic.xaml
```diff
@@ -92,6 +92,7 @@
         <Setter Property="HorizontalOptions" Value="StartAndExpand" />
         <Setter Property="Margin" Value="15,15,15,0" />
         <Setter Property="FontSize" Value="Small" />
+        <Setter Property="IsVisible" Value="{TemplateBinding IsSeekBarVisible}" />
```
jeremyVignelles, line 95: This is a change in the behavior. What if people don't want to link the seekbar and the remaining/elapsed time visibility? I'd be in favor of having separate properties for that, and avoid doing that in this bugfix PR

jeremyVignelles (commented):
LGTM otherwise

src/LibVLCSharp.Forms/Shared/Themes/Generic.xaml
mfkl: We discussed this internally on discord, was my idea. It does not make much sense IMHO to have the time labels (which are relative to the seekbar) without the seekbar. But no strong feelings.

jeremyVignelles: It doesn't make much sense to have a seekbar for infinite/live streams where the duration is unknown. Having the elasped time may be interesting though.
No strong feeling either, just wondering about the behavior change in a bugfix version.

mfkl: > just wondering about the behavior change in a bugfix version.

Sure :-) Happened because Laurent initially opened this PR with this screenshot and that got me wondering.



> for infinite/live streams where the duration is unknown. Having the elasped time may be interesting though.

Fair enough. One property for both labels though to keep it simple. We can revisit this later if needed.

jeremyVignelles: We could also hide the remaining time if the duration is unknown. The behavior is yet to be defined, and I suggest to revert this change and  take the time to discuss more about the wanted new behavior 🙂

mfkl: Sure. /cc @egbakou

egbakou: You mean hide the remaining time label if the MediaPlayer Length=0 ?

jeremyVignelles: Yes, but that's a topic for another PR I guess

egbakou: Okay thanks !
So I guess it's ok for this PR .

jeremyVignelles (approved):

src/LibVLCSharp.Forms/Shared/Themes/Generic.xaml
egbakou: @mfkl  It's done

mfkl:
👍

PR #232, opened 2021-08-25
Make Core.Initialize implicit

mfkl:
Just creating this draft while I think more about it.

It should be documented that:
- if you need a custom location for libvlc, implicit loading won't guess the path for you, you need to call Core.Initialize() yourself with the custom path.
- it should be documented that the first `new LibVLC` can appear slower. That is because the same code now also loads libvlc on first time, while before it was just a regular constructor. Link to guidance to make libvlc load faster (plugins.dat, plugin cherry picking). If still too slow or not acceptable, load libvlc yourself by calling Core.Initialize() _way before_ you call `new LibVLC`.
- No lock is needed because only LibVLC and Equalizer could be created from multiple threads with the `Loaded` value potentially not up to date.
  - You don't want to create multiple LibVLC instance anyway, so you're likely doing something wrong in the first place.
  - If for some reason you want to create multiple Equalizer instances from multiple threads, and run into multithreading issue due to implicit libvlc loading, just load libvlc yourself explicitly beforehand.
  
Need to test on Unity editor/standalone.

### DIFF
--- a/src/LibVLCSharp/Shared/Core/Core.cs
+++ b/src/LibVLCSharp/Shared/Core/Core.cs
@@ -56,8 +56,18 @@ namespace LibVLCSharp.Shared
 #endif
         }
 
+        static bool _libvlcLoaded;
+        static internal bool LibVLCLoaded
+        {
 #if DESKTOP && !NETSTANDARD1_1
-        static bool Loaded => LibvlcHandle != IntPtr.Zero;
+            get => _libvlcLoaded || LibvlcHandle != IntPtr.Zero;
+#else
+            get => _libvlcLoaded;
+#endif
+            set => _libvlcLoaded = value;
+        }
+
+#if DESKTOP && !NETSTANDARD1_1
         static List<(string libvlccore, string libvlc)> ComputeLibVLCSearchPaths()
         {
             var paths = new List<(string, string)>();
@@ -149,7 +159,7 @@ namespace LibVLCSharp.Shared
                     break;
             }
 
-            if (!Loaded)
+            if (!LibVLCLoaded)
             {
                 throw new VLCException("Failed to load required native libraries. " +
                     $"{Environment.NewLine}Have you installed the latest LibVLC package from nuget for your target platform?" +
@@ -157,6 +167,16 @@ namespace LibVLCSharp.Shared
             }
         }
 #endif
+        internal static void EnsureLoaded()
+        {
+            if (LibVLCLoaded)
+            {
+                return;
+            }
+
+            Initialize();
+        }
+
         static bool LoadNativeLibrary(string nativeLibraryPath, out IntPtr handle)
         {
             handle = IntPtr.Zero;
--- a/src/LibVLCSharp/Shared/Internal.cs
+++ b/src/LibVLCSharp/Shared/Internal.cs
@@ -1,5 +1,4 @@
 ﻿using System;
-using System.Runtime.InteropServices;
 
 namespace LibVLCSharp.Shared
 {
--- a/README.md
+++ b/README.md
@@ -253,9 +253,8 @@ For more advanced samples, have a look at [libvlcsharp-samples](https://code.vid
 Feel free to suggest and contribute new samples.
 
 ## Quick API overview
-```csharp
-Core.Initialize();
 
+```csharp
 using var libvlc = new LibVLC(enableDebugLogs: true);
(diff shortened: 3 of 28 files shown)


mfkl:
Should document and update samples. One unintended side effect is that (at least on Windows, maybe other platforms), people might start complaining `new LibVLC` is too slow... that's one of the downsides of *magic* (and lack of plugins.dat for dynamic builds).

jeremyVignelles:
We could at least document prominently the generation of plugins.dat on first run (in the getting started and in the samples). Or we could decide to generate them automatically if you're going that route

robUx4:
How about checking the presence of the plugins.dat and having a mode refusing to init if it's not there (should be the default IMO).

If you have helpers to generate it that shouldn't be too much of a problem for host apps.

Also, unless it's a custom build, we should be able to generate it when building. At least on builds that are built on the target system (may not be the case on Windows).

jeremyVignelles:
I'm not sure plugins.dat is available cross-platform, is it? android? linux?

Shipping a plugins.dat with the nuget package might be an issue though, because of cherry-picking that we allow in our packages. (or we could have a default plugins.dat) if no include/exclude is specified.

Question: Does the plugins.dat include absolute path to dlls or only relative paths?

mfkl:
> How about checking the presence of the plugins.dat and having a mode refusing to init if it's not there (should be the default IMO).

We could output a warning, but a missing plugins.dat shouldn't be a show stopped IMO.

> Also, unless it's a custom build, we should be able to generate it when building. At least on builds that are built on the target system (may not be the case on Windows).

Yes, we could ship the cache gen executable within the nuget package and have it be invoked by MSBuild.

As @jeremyVignelles said, we want to keep support for [plugins cherry picking](https://github.com/mfkl/libvlc-nuget/blob/master/cherry-picking.md) and custom builds.

Currently, this would only work for us on Windows. macOS needs some refactoring to support dynamic builds, and we don't ship libvlc for Linux. iOS and Android are 'static' builds.

> Question: Does the plugins.dat include absolute path to dlls or only relative paths?

I'd assume relative paths.

src/LibVLCSharp/Shared/Core/Core.cs
```diff
@@ -57,7 +59,7 @@ static void Log(string message)
         }
 
 #if (NETFRAMEWORK || NETSTANDARD) && !NETSTANDARD1_1
-        static bool Loaded => LibvlcHandle != IntPtr.Zero;
+        static bool DesktopLoaded => LibvlcHandle != IntPtr.Zero;
```
jeremyVignelles, line 62: `LibraryLoaded` or `NativeLibraryLoaded` or `LibVLCLoaded` instead ?

src/LibVLCSharp/Shared/Internal.cs
```diff
@@ -31,6 +30,8 @@ public abstract class Internal : IDisposable
         /// and performs the release call in native code. It will be called once when the C# instance gets disposed.</param>
         protected Internal(Func<IntPtr> create, Action<IntPtr> release)
         {
+            Core.EnsureLoaded();
```
jeremyVignelles, line 33: I'd say that this is only needed for libVLC construction. I'd put it either in this line:

```cs
        public LibVLC(params string[] options)
            : base(() => Core.EnsureLoaded(); MarshalUtils.CreateWithOptions(PatchOptions(options), Native.LibVLCNew), Native.LibVLCRelease)
```

or inside `MarshalUtils.CreateWithOptions`. What do you think ?

src/LibVLCSharp/Shared/Core/Core.cs
mfkl: Yep, simplified the libvlc loading across targets (there were 2, desktop where we have a native handle and others) and took into account your renaming remark.

src/LibVLCSharp/Shared/Internal.cs
mfkl: Done, would probably be a good idea to document which exceptions can be thrown from `new LibVLC()` now...

jeremyVignelles: Had a discussion recently about that with a friend. Throwing from a constructor might lead to a leak if already-allocated resources are not properly freed... we'll need to be very careful about this...

mfkl: If the `LibVLC` ctor throws, a leak is probably your last problem :) but, yeah, you're right.

PR #244, opened 2021-09-28
CI: add tvOS sample

mfkl:
used to fail the build, don't remember why. Let's try again


mfkl:
Ok I remember now
```
error : No valid tvOS code signing keys found in keychain. You need to request a codesigning certificate from https://developer.apple.com
```

PR #250, opened 2021-11-28
Fixed WPF VideoView foreground window position bug

Yelo420:
### Description of Change ###

For whatever reason, the _bckgnd.LayoutUpdated event only fires inside of a debugger. So lets use the _wndhost.LocationChanged instead. I tested a few cases and everything seems to work fine now.

### Issues Resolved ### 

- fixes #523

### API Changes ###
 
 None

### Platforms Affected ### 

- WPF

The WPF VideoView overlay no longer magically floats around somewhere on the screen

Before
{DE5A8AC6-489E-4B5A-B728-89E2E43B8E8D}

After
{D3F63A78-C087-47BD-A40F-9ECC11A10390}

### Testing Procedure ###
Run LibVlcSharp WPF sample outside of the debugger and move the main window around the screen

### PR Checklist ###

- [x ] Rebased on top of the target branch at time of PR
- [x ] Changes adhere to coding standard

### DIFF
--- a/src/LibVLCSharp.WPF/ForegroundWindow.cs
+++ b/src/LibVLCSharp.WPF/ForegroundWindow.cs
@@ -58,10 +58,11 @@ namespace LibVLCSharp.WPF
         void Background_Unloaded(object sender, RoutedEventArgs e)
         {
             _bckgnd.SizeChanged -= Wndhost_SizeChanged;
-            _bckgnd.LayoutUpdated -= Wndhost_LayoutUpdated;
+            _bckgnd.LayoutUpdated -= RefreshOverlayPosition;
             if (_wndhost != null)
             {
                 _wndhost.Closing -= Wndhost_Closing;
+                _wndhost.LocationChanged -= RefreshOverlayPosition;
             }
 
             Hide();
@@ -84,8 +85,9 @@ namespace LibVLCSharp.WPF
             Owner = _wndhost;
 
             _wndhost.Closing += Wndhost_Closing;
+            _wndhost.LocationChanged += RefreshOverlayPosition;
+            _bckgnd.LayoutUpdated += RefreshOverlayPosition;
             _bckgnd.SizeChanged += Wndhost_SizeChanged;
-            _bckgnd.LayoutUpdated += Wndhost_LayoutUpdated;
 
             try
             {
@@ -100,20 +102,20 @@ namespace LibVLCSharp.WPF
                 Show();
                 _wndhost.Focus();
             }
-            catch(Exception ex)
+            catch (Exception ex)
             {
                 Hide();
                 throw new VLCException("Unable to create WPF Window in VideoView.", ex);
             }
         }
 
-        void Wndhost_LayoutUpdated(object? sender, EventArgs e)
+        void RefreshOverlayPosition(object? sender, EventArgs e)
         {
             if (PresentationSource.FromVisual(_bckgnd) == null)
             {
                 return;
             }
-            
+
             var locationFromScreen = _bckgnd.PointToScreen(_zeroPoint);
             var source = PresentationSource.FromVisual(_wndhost);
             var targetPoints = source.CompositionTarget.TransformFromDevice.Transform(locationFromScreen);


src/LibVLCSharp.WPF/ForegroundWindow.cs
```diff
@@ -58,7 +58,7 @@ void Background_DataContextChanged(object sender, DependencyPropertyChangedEvent
         void Background_Unloaded(object sender, RoutedEventArgs e)
         {
             _bckgnd.SizeChanged -= Wndhost_SizeChanged;
-            _bckgnd.LayoutUpdated -= Wndhost_LayoutUpdated;
+            _bckgnd.LayoutUpdated -= Wndhost_LocationChanged;
```
jeremyVignelles, line 61: You forgot to change the event here

jeremyVignelles (commented):
I don't know ifthose two events are enough to sync the video view. We have a long history of pr trying to fix each other for some edge case. Please check that you don't undo some needed change.

src/LibVLCSharp.WPF/ForegroundWindow.cs
```diff
@@ -84,8 +84,8 @@ void Background_Loaded(object sender, RoutedEventArgs e)
             Owner = _wndhost;
 
             _wndhost.Closing += Wndhost_Closing;
-            _bckgnd.SizeChanged += Wndhost_SizeChanged;
-            _bckgnd.LayoutUpdated += Wndhost_LayoutUpdated;
+            _wndhost.LocationChanged += Wndhost_LocationChanged;
+            _bckgnd.SizeChanged += Wndhost_SizeChanged;
```
mfkl, line 88: nitpick: extra space

mfkl:
This is an exact revert of https://github.com/videolan/libvlcsharp/commit/3fe10a53efe852b3c40fdef063c25737be076c51, which fixed https://code.videolan.org/videolan/LibVLCSharp/-/issues/474.

Yelo420:
@mfkl @jeremyVignelles 
Thanks for your quick reply.
You are right about that. I had not considered that this is actually a revert.

I've committed new solution, maybe this could be it.

Short explanation:
Let's use both Wndhost_LayoutUpdated and Wndhost_LocationChanged. By not firing the Wndhost_LayoutUpdated on window position change, at least outside of a debugger we don't have duplicate function and still every case would be covered. And to avoid firing both functions inside a debugger we have the !Debugger.IsAttached query then.

src/LibVLCSharp.WPF/ForegroundWindow.cs
```diff
@@ -84,8 +85,12 @@ void Background_Loaded(object sender, RoutedEventArgs e)
             Owner = _wndhost;
 
             _wndhost.Closing += Wndhost_Closing;
-            _bckgnd.SizeChanged += Wndhost_SizeChanged;
+            if (!Debugger.IsAttached)
```
jeremyVignelles, line 88: Why is this needed? why not always `on` to avoid a hard-to-spot behavior change depending on whether the debugger is attached?

src/LibVLCSharp.WPF/ForegroundWindow.cs
```diff
@@ -106,8 +111,20 @@ void Background_Loaded(object sender, RoutedEventArgs e)
                 throw new VLCException("Unable to create WPF Window in VideoView.", ex);
             }
         }
-
         void Wndhost_LayoutUpdated(object? sender, EventArgs e)
+        {
```
jeremyVignelles, line 115: Why the code duplication? could it be the same method ? (with a correct name)

src/LibVLCSharp.WPF/ForegroundWindow.cs
```diff
@@ -58,10 +58,11 @@ void Background_DataContextChanged(object sender, DependencyPropertyChangedEvent
         void Background_Unloaded(object sender, RoutedEventArgs e)
         {
             _bckgnd.SizeChanged -= Wndhost_SizeChanged;
-            _bckgnd.LayoutUpdated -= Wndhost_LayoutUpdated;
+            _bckgnd.LayoutUpdated -= Wndhost_LayoutUpdated;
```
mfkl, line 61: Please remove that extra space :)

mfkl:
> I've committed new solution, maybe this could be it.
> 
> Short explanation: Let's use both Wndhost_LayoutUpdated and Wndhost_LocationChanged. By not firing the Wndhost_LayoutUpdated on window position change, at least outside of a debugger we don't have duplicate function and still every case would be covered. And to avoid firing both functions inside a debugger we have the !Debugger.IsAttached query then.

Thanks! It would be great if you could try the minimal repro from https://code.videolan.org/videolan/LibVLCSharp/-/issues/474 to make sure the build from this PR does not break that sample. 

Unfortunately, the original repro sample has been removed by @DengQingFeng (feel free to try this branch if you see this, @DengQingFeng).

jeremyVignelles:
The fact that the LayoutChanged doesn't trigger without the debugger attached might not be the case for every app. It depends on the layout used.
I suspect that when the debugger is attached, VS injects some code in the view, probably a grid or something, to be allowed to display its overlay buttons. When the window moves, that extra code must trigger a LayoutChanged event, but the event is not triggered when that extra code isn't there.
But if you have a more complex app, chances are other things might trigger that LayoutChanged event, and there's no reason to check the debugger.

I'm ok with the fact that the same function can be called twice per move, it should not be very CPU-intensive

src/LibVLCSharp.WPF/ForegroundWindow.cs
Yelo420: you're right. We would create a tricky case. And its only a minimal loss in terms of performance. So it would be ok without.

src/LibVLCSharp.WPF/ForegroundWindow.cs
Yelo420: Here you're right as well. Duplication not needed. Any suggestion for a good function name?

src/LibVLCSharp.WPF/ForegroundWindow.cs
Yelo420: I do 👍

src/LibVLCSharp.WPF/ForegroundWindow.cs
jeremyVignelles: - RefreshPosition
- MoveOverlayWindow
- AlignWindow

Pick you favorite or mix them. The function name should describe what it actually does

Yelo420:
@mfkl @jeremyVignelles 
I've committed once more.
I hope the code layout fits

src/LibVLCSharp.WPF/ForegroundWindow.cs
```diff
@@ -84,8 +85,9 @@ void Background_Loaded(object sender, RoutedEventArgs e)
             Owner = _wndhost;
 
             _wndhost.Closing += Wndhost_Closing;
+            _wndhost.LocationChanged += RefreshOverlayPosition;
```
jeremyVignelles, line 88: Why on _wndhost and not on the background like other events?

Yelo420: Cause _bckgnd has as type FrameworkElement no alternative for LayoutUpdated. So the only other choice to hook on some element is _wndhost.

Yelo420: I've checked: "WindowsFormsHost" The actual type behind "FrameworkElement _bckgnd" also has no alternative. So even a cast would unfortunately not work

mfkl:
Ok, the changes look good now.

afunc233:
This solution does not cope with the VideoView is in the ScrollViewer. any other solution?

jeremyVignelles:
@afunc233 Please make a repro so that we can investigate that, and suggest an event to listen to that could fix the issue

afunc233:
> @afunc233 Please make a repro so that we can investigate that, and suggest an event to listen to that could fix the issue

Wait a moment

afunc233:
image

git repository ：[https://github.com/afunc233/libvlc-wpf-test](github)

jeremyVignelles:
OK so what you're describing is that it works but doesn't "crop" the overlay window to it's container size?
I'm afraid there's nothing we can really do to fix that easily, that's a hack, not a real fix... Do you have any suggestion?

afunc233:
> OK so what you're describing is that it works but doesn't "crop" the overlay window to it's container size? I'm afraid there's nothing we can really do to fix that easily, that's a hack, not a real fix... Do you have any suggestion?

Maybe we can be used  Window.Clip  Property .But I haven't succeeded，Still trying

mfkl:
Is this even related to the PR though? If the issue is already present in
the current nuget builds, then this is off topic, and a proper issue on
gitlab should be created.

jeremyVignelles:
Indeed not related to this PR, let's merge this.

@afunc233 : Please open an issue or a new PR on top of this.

afunc233:
> Indeed not related to this PR, let's merge this.
> 
> @afunc233 : Please open an issue or a new PR on top of this.

ok，thx

mfkl:
Thanks for your contribution @Yelo420.

PR #260, opened 2021-12-09
Added net6.0-windows on WPF and WinForms and workaround MSBuild.Sdk.Extras issue

jeremyVignelles:
### Description of Change ###

- Added net6.0-windows TFM on LibVLCSharp.WinForms
- Added net6.0-windows TFM on LibVLCSharp.WPF
- Added `DESKTOP` constant to replace the ([flaky](https://github.com/novotnyllc/MSBuildSdkExtras/issues/279)) `NETFRAMEWORK || NETSTANDARD || NET6_0`

### Issues Resolved ### 
Users referencing LVS.WinForms in .net6.0 should have a working project now

### API Changes ###
 
None (Fixes missing APIs in 3.6.3)

### Platforms Affected ### 

- Core (all platforms)

### Behavioral/Visual Changes ###

None

### Before/After Screenshots ### 

Not applicable

### PR Checklist ###

- [X] Rebased on top of the target branch at time of PR
- [X] Changes adhere to coding standard

PR #261, opened 2021-12-12
Add WinUI TFM support

drasticactions:
### Description of Change ###

Adds the `net6.0-windows10.0.19041` TFM. This should enable support for using LibVLC in WinUI.

WinUI is more like standard Windows Desktop apps than UWP, so, apart from specific Video Views used for rendering in UI, you should be able to use the standard Windows Desktop DLLs and processes. 

If you try running the video side from within a WinUI app with this, it currently pops out into a new Direct3D window.

### Issues Resolved ### 

- fixes #

### API Changes ###

 
 None

### Platforms Affected ### 

- .NET 6
- WinUI

None

Not applicable

### PR Checklist ###

### DIFF
--- a/src/LibVLCSharp/LibVLCSharp.csproj
+++ b/src/LibVLCSharp/LibVLCSharp.csproj
@@ -31,6 +31,7 @@ LibVLC needs to be installed separately, see VideoLAN.LibVLC.* packages.</Descri
     <TargetFrameworks>netstandard2.1;netstandard2.0;netstandard1.1;net40;net471</TargetFrameworks>
     <TargetFrameworks Condition="!$([MSBuild]::IsOsPlatform('Linux'))">$(TargetFrameworks);monoandroid81;xamarin.ios10;xamarin.mac20</TargetFrameworks>
     <TargetFrameworks Condition="$([MSBuild]::IsOsPlatform('Windows'))">$(TargetFrameworks);uap10.0.18362;net6.0;net6.0-windows;net6.0-ios;net6.0-android;net6.0-tvos;net6.0-macos</TargetFrameworks>
+    <TargetFrameworks Condition="$([MSBuild]::IsOSPlatform('Windows')) and '$(MSBuildRuntimeType)' == 'Full'">$(TargetFrameworks);net6.0-windows10.0.19041</TargetFrameworks>
     <Configurations>Debug;Release;Win32Debug;Win32Release</Configurations>
     <TargetsForTfmSpecificBuildOutput>$(TargetsForTfmSpecificBuildOutput);IncludeAWindow</TargetsForTfmSpecificBuildOutput>
     <RootNamespace>LibVLCSharp</RootNamespace>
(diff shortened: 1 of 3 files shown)


drasticactions:
Also, as a heads up, using `NET6_0` or `NET6_0_WINDOWS` will not work for WinUI support. You need `NET6_0_WINDOWS10_0_19041` for it to work.

jeremyVignelles:
Conflicts with #260

We need a proper WinUI view, not just the TFM to be able to ship it. Please see the open issue about WinUI.

drasticactions:
I'm willing to wait until y'all figure out https://github.com/videolan/libvlcsharp/pull/260, it shouldn't be too hard to rebase after that.

As for a proper WinUI View, I don't think you can do that since you depend on SharpDX. That won't work with WinUI, since it's not WinRT (and it's also a few years old by this point). You still need to use the SwapChainPanel and DirectX, so you would need to convert to a different library, at least for WinUI.

IMO enabling the TFM will at least enable anyone who has an app targeting that specific TFM, since if you don't you can't use LibVLCSharp at all since you can't initialize it. You can technically use that TFM for WinForms, WPF, or a console too (although it's unlikely someone would, but they might want to target specific Windows 10/11 APIs).

If you try running it for WinUI, Audio will be fine. Video works, but will pop out into a new Window until a proper video view is made, but that will only happen with a new library for handling the swap chain panel. I can't really add that, since picking new dependencies for a project should be handled by y'all rather than a contributor like myself.

jeremyVignelles:
Please read the discussion here :
https://code.videolan.org/videolan/LibVLCSharp/-/issues/347

Yes, we will use another binding for DirectX, SharpDX is a good candidate. Yes, we will need to use a SwapChainPanel, but unfortunately, the libvlc option that allows to use a SwapChainPanel is only available in UWP as a hack, and we need libvlc4 to be able to use the new callback-based API.

If I understand correctly, there are two WinUI modes : packaged and full, and it seems to me that your "fix" only works for the full WinUI (the UWP package can't open a popup window IIRC).

My question is : does WinUI fail right now because NuGet doesn't restore the package for the WinUI project (a TFM issue) or does it fail because LibVLCSharp doesn't load libvlc?

If it's the latter, #260 should be enough to fix the issue and it means that NuGet can figure out that it needs to restore the net6.0-windows version. Could you try the #260 branch?

In any case the real WinUI VideoView implementation will need to wait LVS4, and I'm not sure we need to put much effort in just supporting WinUI only in full mode and without a view...

drasticactions:
Right now if you try to add LibVLCSharp to a WinUI project, it will fail because you can't call the Core.Initialize methods, since the TFMs being used don't support the one used by WinUI (net6.0-windows10.0.19041). That goes for _any_ .NET 6 project that happens to use that TFM (but it's more likely the ones using it will be WinUI ones.)

That's why I think at least adding that support for that TFM as a "Desktop" one should be safe and at least allow loading LibVLCSharp. It's going to need to be done regardless if/when WinUI support is fully added, so, IMO, it would be nice to at least load the library for some use cases right now. 


 For your PR,  `net6.0-windows10.0.19041` isn't covered by it so it won't  work. But if you add it for the DESKTOP constant, then I think it should cover it.

jeremyVignelles:
Just tried adding `;net6.0-windows10.0.19041` in the TargetFrameworks in `LibVLCSharp.NetCore.Sample` in #260 and selected that TFM for running through VS, and it worked fine...

This PR is not needed then.

mfkl:
3.6.4 should have targets that will work in WinUI for the core interop. For the WinUI/DirectX integration, there is more work to do.

PR #264, opened 2021-12-28
remove winform host from wpf project

mahdirastegari:
### Description of Change ###

- Remove WinFormsHost from the WPF and create a win32 window and pass it to the MediaPlayer by taking advantage of HwndHost (https://docs.microsoft.com/en-us/dotnet/desktop/wpf/advanced/walkthrough-hosting-a-win32-control-in-wpf)
- Hide this HwndHost when MediaPlayer stopped to don't see the win32's window background and see WPF window background.

### Platforms Affected ### 
- WPF

### DIFF
--- a/src/LibVLCSharp.WPF/LibVLCSharp.WPF.csproj
+++ b/src/LibVLCSharp.WPF/LibVLCSharp.WPF.csproj
@@ -16,7 +16,6 @@ LibVLC needs to be installed separately, see VideoLAN.LibVLC.* packages.
     <RootNamespace>LibVLCSharp.WPF</RootNamespace>
     <PackageId>LibVLCSharp.WPF</PackageId>
     <UseWPF>true</UseWPF>
-    <UseWindowsForms>true</UseWindowsForms>
     <PackageTags>$(PackageTags);wpf</PackageTags>
   </PropertyGroup>
   <!--Override TFMs when building from the LVS.Win32 solution-->
--- a/src/LibVLCSharp.WPF/User32Wrapper.cs
+++ b/src/LibVLCSharp.WPF/User32Wrapper.cs
@@ -1,0 +1,63 @@
+﻿using System;
+using System.Runtime.InteropServices;
+
+namespace LibVLCSharp.WPF
+{
+    internal static class User32Wrapper
+    {
+        public const string LibraryName = "user32.dll";
+
+        public static class EntryPoints
+        {
+            public const string CreateWindow = "CreateWindowEx";
+            public const string DestroyWindow = "DestroyWindow";
+        }
+
+        /// <summary>
+        /// Provide the win32 window stype when call <see cref="EntryPoints.CreateWindow"/>
+        /// <para>See following link: <a href="https://docs.microsoft.com/en-us/windows/win32/winmsg/extended-window-styles" /> </para>
+        /// </summary>
+        [Flags]
+        internal enum ExtendedWindow32Styles : int
+        {
+            /// <summary>
+            /// The window should not be painted until siblings beneath the window (that were created by the same thread) have been painted.
+            /// The window appears transparent because the bits of underlying sibling windows have already been painted.
+            /// </summary>
+            WS_EX_TRANSPARENT = 0x00000020
+        }
+
+        /// <summary>
+        /// Provide the win32 window stype when call <see cref="EntryPoints.CreateWindow"/>
+        /// <para>See following link: <a href="https://docs.microsoft.com/en-us/windows/win32/winmsg/window-styles" /> </para>
+        /// </summary>
+        [Flags]
+        internal enum Window32Styles : int
+        {
+            /// <summary>
+            /// The window is a child window. A window with this style cannot have a menu bar.
+            /// </summary>
+            WS_CHILD = 0x40000000,
+
+            /// <summary>
+            /// The window is initially visible.
+            /// </summary>
+            WS_VISIBLE = 0x10000000
+        }
+
+
+        [DllImport(LibraryName, EntryPoint = EntryPoints.CreateWindow, CharSet = CharSet.Unicode)]
+        internal static extern IntPtr CreateWindowEx(ExtendedWindow32Styles dwExStyle,
+                string lpszClassName,
+                string lpszWindowName,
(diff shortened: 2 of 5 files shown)


src/LibVLCSharp.WPF/LibVLCSharp.WPF.csproj
```diff
@@ -16,7 +16,7 @@ LibVLC needs to be installed separately, see VideoLAN.LibVLC.* packages.
     <RootNamespace>LibVLCSharp.WPF</RootNamespace>
     <PackageId>LibVLCSharp.WPF</PackageId>
     <UseWPF>true</UseWPF>
-    <UseWindowsForms>true</UseWindowsForms>
+    <UseWindowsForms>false</UseWindowsForms>
```
jeremyVignelles, line 19: Just remove the line then

src/LibVLCSharp.WPF/ControlHost.cs
```diff
@@ -0,0 +1,35 @@
...
+using System.Windows.Interop;
+
+namespace LibVLCSharp.WPF
+{
+    /// <summary>
+    /// Provide the UserControl with Handle Pointer in WPF
+    /// <para>This class will create instance of win32 window and hosted it in WPF Framework element</para>
+    /// <remark>As WPF only render the whole window, so all controls over the <see cref="ControlHwndHost"/> will not reneder</remark>
+    /// </summary>
+    public class ControlHwndHost : HwndHost
```
jeremyVignelles, line 17: I'd try to find something less generic. Maybe a VideoHwndHost?

src/LibVLCSharp.WPF/VideoView.cs
```diff
@@ -70,31 +65,73 @@ public override void OnApplyTemplate()
...
+            {
+                _controlHost.Dispatcher.BeginInvoke(new Action(() => _controlHost.Visibility = Visibility.Hidden));
+            }
+        }
+
+        private void MediaPlayer_Playing(object? sender, EventArgs e)
+        {
+            if (_controlHost != null)
+            {
+                _controlHost.Dispatcher.BeginInvoke(new Action(() => _controlHost.Visibility = Visibility.Visible));
```
jeremyVignelles, line 134: I'm not a huge fan of these as this changes the current behavior and is more a matter of personal taste, and that is something that can be done externally.

mahdirastegari: I see your point, but I can see, access to controlHost is kind of hard... but I did implement this code, because I see the issue when I was working with DarkTheme or custom Theme, so I don't want to see the win32's window background as it was the gray.

src/LibVLCSharp.WPF/User32Wrapper.cs
```diff
@@ -0,0 +1,33 @@
+﻿using System;
```
mfkl, line 1: these probably belong to `MarshalUtils.cs`

mahdirastegari: Agreed

mfkl:
Is this a breaking change? Have you tested various airspace scenarios (see commit history on lvs.wpf for example of issues)?

What is the main benefit of these changes?

src/LibVLCSharp.WPF/VideoView.cs
```diff
@@ -4,21 +4,17 @@
...
-using System.Windows.Forms.Integration;
 
 namespace LibVLCSharp.WPF
 {
     /// <summary>
     /// WPF VideoView with databinding for use with LibVLCSharp
     /// </summary>
-    [TemplatePart(Name = PART_PlayerHost, Type = typeof(WindowsFormsHost))]
-    [TemplatePart(Name = PART_PlayerView, Type = typeof(System.Windows.Forms.Panel))]
+    [TemplatePart(Name = PART_PlayerHost, Type = typeof(ControlHwndHost))]
```
jeremyVignelles, line 13: What is preventing us from putting HwndHost here and at `if (Template.FindName(PART_PlayerHost, this) is not ControlHwndHost controlHost)` ?
Do we need something specific from that class ? Can anyone use its own implementation instead of the default one by overriding the template ?

mahdirastegari: I didn't get this, if you mean why did we use ControlHwndHost inside the template? I should tell you, I have followed the previous code, but if you mean why did I remove the Template.FindName(PART_PlayerHost, this) from the public property, because It will call FindName function which I think it doesn't really need it and should call on ApplyTemplate!

mahdirastegari:
@mfkl 
Is there any difference between?
<Grid>
<Grid>
<Grid>
....
<Grid>
<Label Content="I Love You" />
</Grid>
...
</Grid>
</Grid>
</Grid>

with:
<Grid>
<Label Content="I Love You" />
</Grid> ?

As use view, no, but as render time, we couldn't examine that, but definitely, we know that the second one is the best implementation.

src/LibVLCSharp.WPF/User32Wrapper.cs
mahdirastegari: MarshalUtils is internal static class, I don't think it's good idea to make it public.

src/LibVLCSharp.WPF/VideoView.cs
jeremyVignelles: I meant replace this line with:
```cs
[TemplatePart(Name = PART_PlayerHost, Type = typeof(HwndHost))]
```

and replace the line:
```cs
if (Template.FindName(PART_PlayerHost, this) is not ControlHwndHost controlHost)
```

with 
```cs
if (Template.FindName(PART_PlayerHost, this) is not HwndHost windowHost)
```

src/LibVLCSharp.WPF/User32Wrapper.cs
mfkl: We use internalsToVisible for others, not sure it's worth it here though, let me think about it.

mfkl:
Tested click, move and resize events, all working fine.

mfkl (approved):

mfkl:
Thanks! Your contribution will be in the next update.

PR #267, opened 2022-01-11
mediatrack: Use old Id if new StrId API is not used

mfkl:
See https://code.videolan.org/videolan/vlc/-/commit/635ecd1f2b52f976b47b81b6433eec1328d2e880

"All libvlc deprecated functions/members will be removed before VLC 4.0
freeze." 

But the freeze has not happened yet, so we need to keep this working.


mfkl:
This commit will be reverted whenever the 4.0 freeze happens.

PR #269, opened 2022-01-17
Async APIs

mfkl:


mfkl:
Time and position events fire by default without setting them explicitly, so we can discard async signatures for these.

mfkl:
Discarding others as they are likely really fast calls, so probably little added-value. We can always revisit this choice later with customer feedback.

src/LibVLCSharp/Media.cs
```diff
@@ -662,21 +658,22 @@ public bool AddSlave(MediaSlaveType type, uint priority, Uri uri)
...
             cancellationToken.ThrowIfCancellationRequested();
-            ThumbnailerRequest request = default;
 
+            ThumbnailerRequest request = default;
             var tcs = new TaskCompletionSource<Picture>();
-            var cancellationTokenRegistration = cancellationToken.Register(() =>
+
+            var ctr = cancellationToken.Register(() =>
             {
                 ThumbnailGenerated -= OnThumbnailGenerated;
```
jeremyVignelles, line 678: Isn't that the role of unsub() ?

src/LibVLCSharp/MediaPlayer.cs
```diff
@@ -619,10 +620,16 @@ public MediaPlayer(Media media)
...
+        /// Starts playback with Media that is set
         /// If playback was already started, this method has no effect
         /// </summary>
-        /// <returns>true if successful</returns>
+        /// <returns>
+        /// Returns true if the playback will start successfully, false otherwise. 
+        /// This function returns immediately, as it sends a command to the native library. 
+        /// </returns>
+        /// <remarks>
+        /// Playback may not actually have yet started when this function returns.
```
jeremyVignelles, line 631: "See PlayAsync()..."

jeremyVignelles (approved):

mfkl:
Thanks for your comments @jeremyVignelles!

PR #278, opened 2022-03-06
consolidate 3.x packages

jeremyVignelles:
### Description of Change ###

On my computer, I am quite low on disk space, and downloading several versions of the same package, especially libvlc ones, makes things worse.

In this PR, I aligned the versions of the packages and updated the SDK version to 6.0.200


mfkl:
Thanks!

Sidenode: ugh... there is a libvlc regression with one of the media tests which started happening with the latest libvlc 3.x, but I can't pinpoint the exact test or repo locally.

PR #284, opened 2022-03-18
port to eto

rafntor:
Hi, thanks for a nice library!

This PR adds support for the [Eto.Forms](https://github.com/picoe/Eto) platform and I hope it can be included. 
It is made for 3.x branch, I could create a similar for master as well

There are 2 new projects (LibVLCSharp.Eto + LibVLCSharp.Eto.Sample) and they are added to LibVLCSharp.Win32.sln

Buildsystem updates and nuget-generation have not been made and I would need some assistance for that 

Tested on Windows and MacOS ;
mac
win


mfkl:
Hi,

Thank you for the clean contribution! I have no experience with Eto.Forms personally, just heard of it. I wonder how popular it is, but a simple VideoView shouldn't add much maintenance burden.

- Are you planning to use LVS in an Eto.Forms project?
- What are the current solutions for video playback in Eto.Forms-based projects?

samples/LibVLCSharp.Eto.Sample/LibVLCSharp.Eto.Sample.csproj
```diff
@@ -0,0 +1,26 @@
...
+  <!--
+    Set the BuildPlatform property to the Eto platform you wish to build for.
+    The default is the platform you are building on.
+    Valid values: Wpf, Windows, Mac64, XamMac2, Gtk, Direct2D
+-->
+
+  <PropertyGroup>
+    <Title>LibVLCSharp.Eto.Sample</Title>
+    <TargetFramework>net6.0</TargetFramework>
+    <!-- include native libs for compleetnes when publishing selfcontained -->
```
mfkl, line 12: nit: typo

src/LibVLCSharp.Eto/VideoView.cs
```diff
@@ -0,0 +1,98 @@
...
+                MpAttach();
+            }
+        }
+
+        void MpDetach()
+        {
+            if (_mp == null || _mp.NativeReference == IntPtr.Zero)
+                return;
+
+            _mp.Hwnd = IntPtr.Zero;
```
mfkl, line 48: the detach should also do linux/mac.

src/LibVLCSharp.Eto/VideoView.cs
```diff
@@ -0,0 +1,98 @@
...
+        }
+
+        void MpAttach()
+        {
+            if (_mp == null || _mp.NativeReference == IntPtr.Zero)
+                return;
+
+            if (NativeHandle == IntPtr.Zero)
+                throw new InvalidOperationException("VideoView Control must be shown before assigning a MediaPlayer");
```
mfkl, line 57: think that might be a bit strict, we don't currently throw in other integrations in this case. We might revisit that later, but for now this should be aligned with how other platforms behave.

src/LibVLCSharp.Eto/VideoView.cs
```diff
@@ -0,0 +1,98 @@
...
+        /// <param name="disposing"></param>
+        protected override void Dispose(bool disposing)
+        {
+            if (!disposedValue)
+            {
+                if (disposing)
+                {
+                    if (MediaPlayer != null && MediaPlayer.NativeReference != IntPtr.Zero)
+                    {
+                        MediaPlayer.Hwnd = IntPtr.Zero;
```
mfkl, line 91: Ditto, platform-specific detach needed here.

mfkl:
You'll need to rebase your branch on latest `3.x` to include the CI fix https://github.com/videolan/libvlcsharp/pull/285

mfkl:
Can you add controls easily on top of the video? Or are there going to be "airspace issues"?

rafntor:
Rebased, and addressed the other issues you found ;-)

- I also dont know how much used Eto.Forms is, but I like the approach and have tried to adapt relevant resources (like libvlc) to be usable for the platform. But it is active and issues are dealt with so I hope it keeps on going.
- I have no immediate plans for using the library- I just realized it was low-hanging-fruit to connect these frameworks and think it makes a good addition to both projects. I also dont know of any alternatives 
- Adding controls on top should be possible but I have not tried ; Eto has a flexible layout-system but stacking controls over each other would require the [PixelLayout](https://github.com/picoe/Eto/wiki/PixelLayout) which is only for special use cases. I may give it a shot to check ;-)

rafntor:
That did not work unfortunately , this code ; 

```csharp
videoView1.Content = new Button { Text = "button5", Size = new Size(60, 20) };
layout.SizeChanged += (o, e) =>
{
    videoView1.Size = layout.Size;
    var p = Point.Empty;
    p.X = layout.Width - b2.Width;
    layout.Move(b2, p);
    p.Y += layout.Height - b3.Height;
    layout.Move(b3, p);
    p.X = 0;
    layout.Move(b4, p);
};
videoView1.MouseUp += (o, e) =>
{
    var media = new Media(_libVLC, new Uri("http://commondatastorage.googleapis.com/gtv-videos-bucket/sample/BigBuckBunny.mp4"));
    _mp.Play(media);
    media.Dispose();
};
```

.. shows like this 
image

but as soon as I click to start the movie the buttons are gone:
image

It should not be a big problem I think, it is a special situation(?)

mfkl:
> It should not be a big problem I think, it is a special situation(?)

Yes, not a problem for initial support of Eto/LibVLCSharp, we can merge your PR as-is. But I expect users will soon after open tickets such as "I'm trying to build a video player with buttons and controls on top of the video view and it won't work".

This is because vlc draws on the native window handle and most UI frameworks don't handle multiple renders on the same native window handle. We've had to deal with this, also known as "airspace issues", with WPF and Avalonia as well, see https://code.videolan.org/videolan/LibVLCSharp/-/issues?label_name%5B%5D=airspace.

Usual good-enough fix, is to have a transparent window on top of the video to put controls and stuff on. But we can see about this at a later time :-)

rafntor:
Pehaps @cwensley knows if it is possible to deal with "airspace issues" as described above?

jeremyVignelles:
> This is because vlc draws on the native window handle and most UI frameworks don't handle multiple renders on the same native window handle. We've had to deal with this, also known as "airspace issues", with WPF and Avalonia as well, see https://code.videolan.org/videolan/LibVLCSharp/-/issues?scope=all&state=all&label_name%5B%5D=airspace.

Sorry to interrupt, but you are talking about two different things:
- Airspace issues found in WPF and Avalonia, which is because a HWND context is put inside a window with a different rendering mechanism, allowing embedding stuff, but preventing said rendering mechanism to draw over it
- Using the window's HWND as a source for libvlc, which takes over the whole rendering mechanism, and prevent users to put any other control, including buttons to play it, which is... annoying 😄 

If we are in the former case, sure, we can deal with it, but if we are in the second one, we can't merge as-is.

@rafntor : Could you try to put controls under the video and see if that works ? If the video still takes the full window, then that's because you're using the window's HWND and not the control's HWND as in WinForms (which may or may not be possible in Eto.Forms)

jeremyVignelles:
@rafntor :
`videoView1.Content = new Button { Text = "button5", Size = new Size(60, 20) };`
This is not a good thing to do as obviously, the video content will be overwritten, but what about a button at the same level as the video view and placed above the video ?

samples/LibVLCSharp.Eto.Sample/LibVLCSharp.Eto.Sample.csproj
```diff
@@ -0,0 +1,26 @@
...
+  <!--
+    Set the BuildPlatform property to the Eto platform you wish to build for.
+    The default is the platform you are building on.
+    Valid values: Wpf, Windows, Mac64, XamMac2, Gtk, Direct2D
+-->
+
+  <PropertyGroup>
+    <Title>LibVLCSharp.Eto.Sample</Title>
+    <TargetFramework>net6.0</TargetFramework>
+    <!-- include native libs for completenes when publishing selfcontained -->
```
mfkl, line 12:
```suggestion
    <!-- include native libs for completeness when publishing self-contained -->
```

src/LibVLCSharp.Eto/VideoView.cs
```diff
@@ -0,0 +1,92 @@
...
+                MpAttach();
+            }
+        }
+
+        void MpDetach()
+        {
+            if (_mp == null || _mp.NativeReference == IntPtr.Zero)
+                return;
+
+            _mp.Hwnd = _mp.NsObject = IntPtr.Zero;
```
mfkl, line 48: It may be harmless but we want to avoid all unnecessary native calls to libvlc. On mac, this calls into code like

```
void libvlc_media_player_set_nsobject( libvlc_media_player_t *p_mi,
                                        void * drawable )
{
    assert (p_mi != NULL);
#ifdef __APPLE__
    var_SetString (p_mi, "dec-dev", "any");
    var_SetString (p_mi, "vout", "any");
    var_SetString (p_mi, "window", "any");
    var_SetAddress (p_mi, "drawable-nsobject", drawable);
#else
    (void)drawable;
    libvlc_printerr ("can't set nsobject: APPLE build required");
    assert(false);
    var_SetString (p_mi, "window", "none");
#endif
}
```

`if (RuntimeInformation.IsOSPlatform(OSPlatform.X))` should be used here as well to avoid calling platform-specific function on any platform.

samples/LibVLCSharp.Eto.Sample/MainForm.cs
```diff
@@ -0,0 +1,43 @@
...
+    using global::Eto.Forms;
+    using global::Eto.Drawing;
+    using LibVLCSharp.Shared;
+    using System;
+
+    public class MainForm : Form
+    {
+        LibVLC _libVLC;
+        MediaPlayer _mp;
+        VideoView videoView1;
```
mfkl, line 13:
```suggestion
        VideoView videoView;
```

samples/LibVLCSharp.Eto.Sample/MainForm.cs
```diff
@@ -0,0 +1,43 @@
...
+            Title = $"Eto.Forms - { Platform } - { Environment.OSVersion }";
+            MinimumSize = new Size(16, 9) * 10;
+            Size = MinimumSize * 6;
+
+            Core.Initialize();
+
+            _libVLC = new LibVLC();
+            _mp = new MediaPlayer(_libVLC);
+            
+            Content = videoView1 = new VideoView();
```
mfkl, line 25:
```suggestion
            Content = videoView = new VideoView();
```

samples/LibVLCSharp.Eto.Sample/MainForm.cs
```diff
@@ -0,0 +1,43 @@
...
+        }
+        protected override void OnShown(EventArgs e)
+        {
+            videoView1.MediaPlayer = _mp;
```
mfkl, line 29:
```suggestion
            videoView.MediaPlayer = _mp;
```

samples/LibVLCSharp.Eto.Sample/MainForm.cs
```diff
@@ -0,0 +1,43 @@
...
+            var media = new Media(_libVLC, new Uri("http://commondatastorage.googleapis.com/gtv-videos-bucket/sample/BigBuckBunny.mp4"));
+            _mp.Play(media);
+            media.Dispose();
+            base.OnShown(e);
+        }
+        protected override void OnClosed(EventArgs e)
+        {
+            _mp.Stop();
```
mfkl, line 37: we may want to dispose of the video view here?

mfkl:
> There are 2 new projects (LibVLCSharp.Eto + LibVLCSharp.Eto.Sample) and they are added to LibVLCSharp.Win32.sln

The CI only builds `LibVLCSharp.sln` and `LibVLCSharp.Mac.sln`, I think they should be added to `LibVLCSharp.sln` (at least) instead. `LibVLCSharp.Win32.sln` is just for dev/testing.

rafntor:
@mfkl I believe latest commits covers your concerns, let me know if you spot something else

rafntor:
@jeremyVignelles - I think it is the controls handle that gets passed to vlc (not main-window). I will try to put up a demo to 'prove' it :wink:
The 4 corner-buttons in first pic here https://github.com/videolan/libvlcsharp/pull/284#issuecomment-1074270485 are placed above the videoview-control (Eto.Forms does not provide any z-order function, ordering is done on a fifo-basis how it is added to a PixelLayout list)

rafntor:
This is better ;
image

I changed the Eto.Platform from Wpf (default) to Winforms (notice window titles in the screenshorts)
So - it seems the EtoWpf.Platform provides bad handles for the Controls ; that is an Eto-bug then if I am not mistaken

cwensley:
@rafntor the only way to get rid of the airspace issue in WPF is to get VLC to render to an offscreen buffer then draw that to the screen, instead of rendering to a specific HWND.  This would mean writing some WPF-specific code to do that, perhaps using [InteropBitmap](https://docs.microsoft.com/en-us/dotnet/api/system.windows.interop.interopbitmap?view=windowsdesktop-6.0).  Or perhaps we could create a general purpose control for this scenario in Eto.

Also, WPF only has a single HWND for the entire window, and you'd typically need to use [HwndHost](https://docs.microsoft.com/en-us/dotnet/api/system.windows.interop.hwndhost?view=windowsdesktop-6.0) to create an HWND for a specific control.  It also has airspace issues, however.

jeremyVignelles:
Oh, so Eto uses either WinForms or WPF as a backend ?
Is the end-user responsible for such decision ?

As cwensley explained, in WPF, you don't have a different HWND for each control, as opposed to WinForms.

The WinForms implementation is trivial, you've already done it, but the WPF one is a bit trickier (see the WPF control that we have). Having airspace issues in a WPF implementation of Eto is acceptable for this PR, the overlay window hack could come later, if necessary at all.

rafntor:
Thanks for chiming in ;-) This was all unknown to me and that makes it not so low-hanging after all unfortunately. Platform-dependent assemblies was not a direction I like

rafntor:
@jeremyVignelles  - Yes , Eto allows user to select what to build for. Any project can be built something like this;
- dotnet build -p:BuildPlatform Windows
- dotnet build -p:BuildPlatform Wpf
- dotnet build -p:BuildPlatform Gtk
- dotnet build -p:BuildPlatform Mac64
- dotnet build -p:BuildPlatform XamMac

So 2 variations for Windows & MacOS 
This PR naively aimed to target any and all of them since it only builds on functionality common across all platforms, too naively as we no see unfortunately :wink:

cwensley:
@rafntor makes sense, I've added https://github.com/picoe/Eto/issues/2167 and https://github.com/picoe/Eto/issues/2168 to  address those issues.

rafntor:
Great! A new control in Eto like that would make it trivial to later update and get proper Wpf behavior for libVlc

jeremyVignelles:
> dotnet build -p:BuildPlatform Gtk

Does this platform work on Windows and mac as well, or just linux ?

I'll need to try that one for the Linux port.

https://github.com/picoe/Eto/issues/2167 is a blocker for LibVlcSharp on Eto.Forms.Wpf . This platform can be explicitely left unimplemented for the time being.

https://github.com/picoe/Eto/issues/2168 will likely have perfs issue for the video as there will be lots of copies between the GPU (where the frames are decoded) and the RAM, where your bitmap will live (before being rendered back to the GPU).

PS : If both of you want to talk more about those issues, feel free to come over at our discord server 🙂

mfkl:
Thank you for your contribution @rafntor! This will be part of the next incoming release.

rafntor:
Great - thanks for all the support 👍

mfkl:
https://www.nuget.org/packages/LibVLCSharp.Eto/

rafntor:
@jeremyVignelles  were you able to run the Gtk-version?

jeremyVignelles:
Didn't had time to check, mfkl merged before I could give it a try. I will try to find the time in the next days.

samples/LibVLCSharp.Eto.Sample/Program.cs
```diff
@@ -0,0 +1,22 @@
...
+        [STAThread]
+        public static void Main(string[] args)
+        {
+            try
+            {
+                new Application(global::Eto.Platform.Detect).Run(new MainForm());
+            }
+            catch (Exception ex)
+            {
+                MessageBox.Show(ex.Message, "Exception !", MessageBoxType.Error);
```
jeremyVignelles, line 18: MessageBox.Show throws here if the Application was not created yet (like `new Application(global::Eto.Platforms.Gtk2)` when I did not install the corresponding package.

jeremyVignelles:
It doesn't work as-is for Gtk on linux.

First, I got this exception:

```
Marshaling show signal
Exception in Gtk# callback delegate
  Note: Applications can use GLib.ExceptionManager.UnhandledException to handle the exception.
System.Reflection.TargetInvocationException: Exception has been thrown by the target of an invocation.
 ---> System.OverflowException: Arithmetic operation resulted in an overflow.
   at System.IntPtr.op_Explicit(IntPtr value)
   at LibVLCSharp.Eto.VideoView.MpAttach(IntPtr handle) in /data/projects/dotnet/libvlcsharp/src/LibVLCSharp.Eto/VideoView.cs:line 72
   at LibVLCSharp.Eto.VideoView.set_MediaPlayer(MediaPlayer value) in /data/projects/dotnet/libvlcsharp/src/LibVLCSharp.Eto/VideoView.cs:line 53
```
which relates to this line:

```cs
_mp.XWindow = (uint)handle;
```
It seems like in GTK#, the handle is not a xid directly, but a gtk widget handle. As we see here for mono/gtksharp 2, we need to call native functions to get the window identifiers :

https://github.com/videolan/libvlcsharp/blob/65a5e8011904045040aa400e518b08de2d79119e/src/LibVLCSharp.GTK/VideoView.cs#L88-L98

Of course, everything changed between GTK 2 and 3, and the functions are not the same, so I came up with:

```cs
_mp.XWindow = Native.gdk_x11_window_get_xid(Native.gtk_widget_get_window(handle));
```

Loading the correct functions is out of scope of this very long message, but it's hard to do in itself because GTK# doesn't expose it, and even if they did, we wouldn't want to take a hard dependency on GTK#...

This... still fails to this message:

```
(LibVLCSharp.Eto.Sample:47118): Gdk-WARNING **: 23:13:59.768: ../../../../../gdk/x11/gdkwindow-x11.c:5653 drawable is not a native X11 window
```

It turns out GTK3 defaults to Wayland by default, which VLC doesn't support. I run the application again with:

```sh
GDK_BACKEND=x11 dotnet run
```

And now, I have my application running as expected. But I'm unsure whether `Native.gtk_widget_get_window` gets a "window" for that widget, or if it finds the parent window, which would be problematic for VLC as described in a previous message.

I tried to edit the sample like so:

```cs
            
            videoView = new VideoView();

            var layout = new PixelLayout();
            layout.Add(videoView, Point.Empty);
            layout.Add(new Button() { Text = "Hello", Width = 150, Height=50 }, new Point(10, 10));

            Content = layout;
```

But the player now crashes without any explanation.
gdb gives me this explanation:

```
#0  ___pthread_mutex_lock (mutex=0x0) at pthread_mutex_lock.c:76
#1  0x00007ffff080778f in XrmQGetResource () from /lib/x86_64-linux-gnu/libX11.so.6
#2  0x00007ffff07e8f9a in XGetDefault () from /lib/x86_64-linux-gnu/libX11.so.6
#3  0x00007ffff071f4d8 in ?? () from /lib/x86_64-linux-gnu/libcairo.so.2
#4  0x00007ffff0722264 in ?? () from /lib/x86_64-linux-gnu/libcairo.so.2
#5  0x00007ffff06f5aa9 in cairo_surface_get_font_options () from /lib/x86_64-linux-gnu/libcairo.so.2
#6  0x00007ffff06aa982 in ?? () from /lib/x86_64-linux-gnu/libcairo.so.2
#7  0x00007ffff06ae8b3 in ?? () from /lib/x86_64-linux-gnu/libcairo.so.2
#8  0x00007ffff07056fb in cairo_show_glyphs () from /lib/x86_64-linux-gnu/libcairo.so.2
#9  0x00007ffff090ccb3 in ?? () from /lib/x86_64-linux-gnu/libpangocairo-1.0.so.0
#10 0x00007ffff090cf35 in ?? () from /lib/x86_64-linux-gnu/libpangocairo-1.0.so.0
#11 0x00007ffff02897c3 in pango_renderer_draw_glyph_item () from /lib/x86_64-linux-gnu/libpango-1.0.so.0
#12 0x00007ffff0291177 in pango_renderer_draw_layout_line () from /lib/x86_64-linux-gnu/libpango-1.0.so.0
#13 0x00007ffff0291805 in pango_renderer_draw_layout () from /lib/x86_64-linux-gnu/libpango-1.0.so.0
#14 0x00007ffff090da92 in pango_cairo_show_layout () from /lib/x86_64-linux-gnu/libpangocairo-1.0.so.0
#15 0x00007ffff0ce3190 in gtk_render_layout () from /lib/x86_64-linux-gnu/libgtk-3.so.0
#16 0x00007ffff0c529f0 in ?? () from /lib/x86_64-linux-gnu/libgtk-3.so.0
#17 0x00007ffff0b91131 in ?? () from /lib/x86_64-linux-gnu/libgtk-3.so.0
#18 0x00007ffff0ba56a2 in ?? () from /lib/x86_64-linux-gnu/libgtk-3.so.0
#19 0x00007ffff0c4c235 in ?? () from /lib/x86_64-linux-gnu/libgtk-3.so.0
#20 0x00007ffff0b2e683 in ?? () from /lib/x86_64-linux-gnu/libgtk-3.so.0
#21 0x00007ffff0db89b4 in ?? () from /lib/x86_64-linux-gnu/libgtk-3.so.0
#22 0x00007ffff0b99695 in gtk_container_propagate_draw () from /lib/x86_64-linux-gnu/libgtk-3.so.0
#23 0x00007ffff0b9979d in ?? () from /lib/x86_64-linux-gnu/libgtk-3.so.0
#24 0x00007ffff0b55271 in ?? () from /lib/x86_64-linux-gnu/libgtk-3.so.0
#25 0x00007ffff0b91131 in ?? () from /lib/x86_64-linux-gnu/libgtk-3.so.0
#26 0x00007ffff0ba56a2 in ?? () from /lib/x86_64-linux-gnu/libgtk-3.so.0
#27 0x00007ffff0b525e5 in ?? () from /lib/x86_64-linux-gnu/libgtk-3.so.0
#28 0x00007ffff0db89b4 in ?? () from /lib/x86_64-linux-gnu/libgtk-3.so.0
#29 0x00007ffff0b99695 in gtk_container_propagate_draw () from /lib/x86_64-linux-gnu/libgtk-3.so.0
#30 0x00007ffff0b9979d in ?? () from /lib/x86_64-linux-gnu/libgtk-3.so.0
#31 0x00007ffff0b3b451 in ?? () from /lib/x86_64-linux-gnu/libgtk-3.so.0
#32 0x00007ffff0b91131 in ?? () from /lib/x86_64-linux-gnu/libgtk-3.so.0
#33 0x00007ffff0ba56a2 in ?? () from /lib/x86_64-linux-gnu/libgtk-3.so.0
#34 0x00007ffff0b3bcf5 in ?? () from /lib/x86_64-linux-gnu/libgtk-3.so.0
#35 0x00007ffff0db89b4 in ?? () from /lib/x86_64-linux-gnu/libgtk-3.so.0
#36 0x00007ffff0b99695 in gtk_container_propagate_draw () from /lib/x86_64-linux-gnu/libgtk-3.so.0
#37 0x00007ffff0c0b761 in ?? () from /lib/x86_64-linux-gnu/libgtk-3.so.0
#38 0x00007ffff0db89b4 in ?? () from /lib/x86_64-linux-gnu/libgtk-3.so.0
#39 0x00007ffff0b99695 in gtk_container_propagate_draw () from /lib/x86_64-linux-gnu/libgtk-3.so.0
#40 0x00007ffff0b9979d in ?? () from /lib/x86_64-linux-gnu/libgtk-3.so.0
#41 0x00007ffff0db89b4 in ?? () from /lib/x86_64-linux-gnu/libgtk-3.so.0
#42 0x00007ffff0b99695 in gtk_container_propagate_draw () from /lib/x86_64-linux-gnu/libgtk-3.so.0
#43 0x00007ffff0b9979d in ?? () from /lib/x86_64-linux-gnu/libgtk-3.so.0
#44 0x00007ffff0b3b451 in ?? () from /lib/x86_64-linux-gnu/libgtk-3.so.0
#45 0x00007ffff0b91131 in ?? () from /lib/x86_64-linux-gnu/libgtk-3.so.0
#46 0x00007ffff0ba56a2 in ?? () from /lib/x86_64-linux-gnu/libgtk-3.so.0
#47 0x00007ffff0b3bcf5 in ?? () from /lib/x86_64-linux-gnu/libgtk-3.so.0
#48 0x00007ffff0db89b4 in ?? () from /lib/x86_64-linux-gnu/libgtk-3.so.0
#49 0x00007ffff0b99695 in gtk_container_propagate_draw () from /lib/x86_64-linux-gnu/libgtk-3.so.0
#50 0x00007ffff0b9979d in ?? () from /lib/x86_64-linux-gnu/libgtk-3.so.0
#51 0x00007ffff0b3b451 in ?? () from /lib/x86_64-linux-gnu/libgtk-3.so.0
```

Which sounds awfully bad.

If you have pointers or ideas of things I could try, feel free to come and chat on [our Discord server](https://discord.gg/3h3K3JF)

rafntor:
The `NativeHandle` we use I find here ; [Widget.NativeHandle](https://github.com/picoe/Eto/blob/a4d371382a80620870cdedfb58cca3a7cd869fb2/src/Eto.Gtk/Forms/GtkControl.cs#L91)
Which I think is [`GLibSharp.Glib.Object.Handle`](https://github.com/GtkSharp/GtkSharp/blob/7995f132e19e18b8dae94359246540f059e53a86/Source/Libs/GLibSharp/Object.cs#L742) and assigned [here?](https://github.com/GtkSharp/GtkSharp/blob/7995f132e19e18b8dae94359246540f059e53a86/Source/Libs/GLibSharp/Object.cs#L717)
But I have no knowledge of what goes on in GtkSharp and what flavors of handles might be involved unfortunately, can only hope @cwensley has an tip to something that can be attempted

jeremyVignelles:
get_window/get_xid looks like the way to go with the gtk platform on linux x11, but it looks like we have to do a lot of if/else for specific platforms if we want gtk support

rafntor:
Possibly the control in https://github.com/picoe/Eto/issues/2167 could provide such a handle on Gtk - that would make life easier ?
Possibly the theorized control could have a `GetHandle(type)` function that takes a `enum HandleType` argument with relevant types that covers the needs for whatever handle-flavor is needed .. just throwing up wild suggestions here ;-)

PR #296, opened 2022-05-24
sample: add support to latest Avalonia

mfkl:
### Description of Change ###

Fix support to Avalonia 0.10.14

### Issues Resolved ### 

- fixes https://code.videolan.org/videolan/LibVLCSharp/-/issues/559
https://github.com/AvaloniaUI/Avalonia/issues/8004

### API Changes ###

 None

### Platforms Affected ### 
- Windows


robUx4:
The Vista support won't be true for VLC 4.0

mfkl:
Good catch, thanks.

PR #298, opened 2022-06-13
WinForms: Use provided handle management APIs

mfkl:
### Description of Change ###

Use provided WinForms APIs.

### Issues Resolved ### 

- fixes https://code.videolan.org/videolan/LibVLCSharp/-/issues/521

### API Changes ###

 Removed:
 - ISupportInitialize

### Platforms Affected ### 

- WinForms

None

Not applicable


jeremyVignelles:
I forgot about my suggestion sorry.
Since then, in my player, I found that the "DestroyHandle" is not always called. IIRC, it was when the application is shutting down.
I put some code in the Dispose() method, but I'm not sure I've been able to fix this... It was long ago... 😅 
Do you know if it works well in your case ?

mfkl:
https://stackoverflow.com/a/43560/4064749 some info here.

PR #305, opened 2022-08-24
mediaplayer: Add record binding

mfkl:
The event is not currently firing, apparently something is incorrect on our side.


mfkl:
This could benefit some unit test, but tests don't run currently (even locally) https://github.com/videolan/libvlcsharp/pull/150 first, so, later.

PR #306, opened 2022-08-29
mediaplayer: fix vmem cleanup callback crash

mfkl:


src/LibVLCSharp/Shared/MediaPlayer.cs
```diff
@@ -1828,16 +1827,6 @@ private static uint VideoFormatCallback(ref IntPtr opaque, IntPtr chroma, ref ui
             return 0;
         }
 
-        [MonoPInvokeCallback(typeof(LibVLCVideoCleanupCb))]
```
jeremyVignelles, line 1831: I don't understand why you remove a PInvoke callback and replace it with the user-provided callback.
If I remember correctly, those callbacks were used when used with mono on iOS, because you couldn't pass non-static callbacks to a native library.

Anyway, it seems weird to remove it here and keep the other PInvoke callbacks, and if it actually fixes the issue, that's more likely "by chance" (a.k.a not failing hard enough to reveal the crash) than because of a really proper fix. Or am I missing something ?

jeremyVignelles: Also, by doing that, you don't pass the `_videoUserData` that the user has provided with the method

mfkl:
```
static void Close(vlc_object_t *object)
{
    vout_display_t *vd = (vout_display_t *)object;
    vout_display_sys_t *sys = vd->sys;

    if (sys->cleanup)
        sys->cleanup(sys->opaque);
    if (sys->pool)
        picture_pool_Release(sys->pool);
    free(sys);
}
```

The static callbacks version requires a "valid" mediaplayer that you can cast into from native code. Unlike the audio callbacks, the video callbacks don't have a Stop function, only a Close one (which performs both cleanup + free).

Since this is the only callback affected (e.g. that crashes the CLR) and that the previous version worked fine, my (educated) guess is that the `free(sys)` was running before the CLR did its cast in the static callback version of the code.

https://github.com/videolan/vlc/blob/5f94c5b72f3a2e4b6d6d49e23740b3b6996ec944/modules/video_output/vmem.c#L224-L234

mfkl: > Also, by doing that, you don't pass the _videoUserData that the user has provided with the method

It doesn't matter. LVS is the one using that internally mostly. Here we don't. If some user really needs to use this (corner of corner case), they can re-implement this function call (and reach out to share their use case :D)

mfkl: > those callbacks were used when used with mono on iOS, because you couldn't pass non-static callbacks to a native library.

Hmm true. Not sure the current way is problematic though, will have to test somehow.

mfkl: It can't be worst than now at least (which is a CLR crash) xD

jeremyVignelles: > The static callbacks version requires a "valid" mediaplayer that you can cast into from native code.

It requires a valid .net MediaPlayer instance, but does the native handle need to be valid to access the _videoCleanupCb field ?

Where does the CLR crash ?

mfkl: > but does the native handle need to be valid to access the _videoCleanupCb field ?

It apparently does.

>Where does the CLR crash ?

When we try to cast the opaque pointer back to a mediaplayer instance.
https://github.com/videolan/libvlcsharp/blob/b32538d3c2fa2b84a5d25d41badd1e743b57e470/src/LibVLCSharp/Shared/Helpers/MarshalUtils.cs#L665. The struct containing the opaque pointer has been free'd and the CLR doesn't like that.

PR #307, opened 2022-08-29
Uwp capabilities

robUx4:
### Description of Change ###

Add the capabilities that a regular UWP app using libvlcsharp can use. This is a better reference than the old vlc-winrt repository.

See the list of capabilities available: https://docs.microsoft.com/en-us/windows/uwp/files/file-access-permissions#capabilities-for-accessing-other-locations

### Platforms Affected ### 

- UWP


mfkl:
Thanks!

PR #319, opened 2022-12-14
Update ForegroundWindow.cs

MakishimuAkuma:
### Issues Resolved ### 

- fixes #Fixes application crash when using [wpfui](https://github.com/lepoco/wpfui) when switching pages
- fixes #Unable to create WPF Window in VideoView.

### API Changes ###

 
 None

### Platforms Affected ### 

- WPF

None

https://user-images.githubusercontent.com/61804392/208105795-e11a5056-99c7-43f3-8aec-51d0be6e878a.mp4

https://user-images.githubusercontent.com/61804392/208112968-f4a0e3df-499f-44bc-9ab2-ef08bf0b7adc.mp4

### PR Checklist ###

### DIFF
--- a/src/LibVLCSharp.WPF/ForegroundWindow.cs
+++ b/src/LibVLCSharp.WPF/ForegroundWindow.cs
@@ -81,6 +81,11 @@ namespace LibVLCSharp.WPF
             {
                 return;
             }
+            
+            if (PresentationSource.FromVisual(_bckgnd) == null)
+            {
+                return;
+            }
 
             Owner = _wndhost;
 


jeremyVignelles:
Could you explain what the issue is ? What does this solve and why should we import this in our library?
Do you have any code that demonstrates the issue ?

MakishimuAkuma:
After adding the problem is solved

```
void Background_Loaded(object? sender, RoutedEventArgs e)
        {
            if (_wndhost != null && IsVisible)
            {
                return;
            }

            _wndhost = GetWindow(_bckgnd);
            Trace.Assert(_wndhost != null);
            if (_wndhost == null)
            {
                return;
            }

            if (PresentationSource.FromVisual(_bckgnd) == null)
            {
                return;
            }

            Owner = _wndhost;

            _wndhost.Closing += Wndhost_Closing;
            _wndhost.LocationChanged += RefreshOverlayPosition;
            _bckgnd.LayoutUpdated += RefreshOverlayPosition;
            _bckgnd.SizeChanged += Wndhost_SizeChanged;

            try
            {
                var locationFromScreen = _bckgnd.PointToScreen(_zeroPoint);
                var source = PresentationSource.FromVisual(_wndhost);
                var targetPoints = source.CompositionTarget.TransformFromDevice.Transform(locationFromScreen);
                Left = targetPoints.X;
                Top = targetPoints.Y;
                var size = new Point(_bckgnd.ActualWidth, _bckgnd.ActualHeight);
                Height = size.Y;
                Width = size.X;
                Show();
                _wndhost.Focus();
            }
            catch (Exception ex)
            {
                Hide();
                throw new VLCException("Unable to create WPF Window in VideoView.", ex);
            }
        }
```

mfkl:
```
throw new VLCException("Unable to create WPF Window in VideoView.", ex);
```
Could you share the value of the inner exception `exe` please?

I agree the code should not throw and your fix seems OK, but I'd like to understand what WPF UI is doing to trigger this.

Also, in your video that shows "after", I understand there is no more exception from the videoview, but it seems the video is not playing either. Did you manage to make it play in a WPF UI app?

Thanks!

MakishimuAkuma:
I'm sorry, the video plays without any problems.
The problem was only with the application crashing when changing the page.

mfkl:
> Could you share the value of the inner exception exe please?

We are still interested in the inner exception message and stack, if you have it. :)

Can you check that your changes does not break existing samples please?
Namely
- https://github.com/videolan/libvlcsharp/tree/3.x/samples/Forms/LibVLCSharp.Forms.WPF.Sample
- https://github.com/videolan/libvlcsharp/tree/3.x/samples/LibVLCSharp.WPF.Sample

Thanks!

mfkl:
It would be nice to have a sample we can clone and run to see the issue. Additionally, a comment explaining in which cases `PresentationSource.FromVisual(_bckgnd)` can be null would be helpful to better understand the need for this check.

mfkl:
This looks similar to https://github.com/videolan/libvlcsharp/commit/ec64571a37fcf4a164a185f4177ac5e7f6ea7496. There might be better ways to handle this (https://stackoverflow.com/questions/11204251/presentationsource-fromvisualthis-returns-null-value-in-wpf) but this could be good enough to merge for now.

mfkl:
@MakishimuAkuma Could you please share a sample that reproduce the issue?

mfkl:
Closing for inactivity.

PR #327, opened 2023-06-21
Fix a typo in MediaPlayer property name

nightfog-git:
### Description of Change ###
Fix a typo in ProgramScrambled property name. Master branch is also affected.

### API Changes ###
Changed:
 - MediaPlayer.ProgramScambled => MediaPlayer.ProgramScrambled

### Platforms Affected ### 
- Core (all platforms)


mfkl:
How about putting this under an [Obsolete] property and adding the new property side by side? This way we don't break existing user code and they have a warning that it will break in the next major version.

mfkl:
merged on master https://github.com/videolan/libvlcsharp/commit/e7917c17b7447566885825e12296689dd1509ab6

PR #341, opened 2024-04-15
mediaplayer: add video fit mode

mfkl:


src/LibVLCSharp/MediaPlayer.cs
```diff
@@ -3596,4 +3613,35 @@ public enum AudioOutputMixmode
...
+
+    /// <summary>
+    /// Enumeration of values used to set the video fitting inside the display area.
+    /// </summary>
+    public enum VideoFitMode
+    {
+        /// <summary>
+        /// Explicit zoom set by <see cref="MediaPlayer.Scale"/>
+        /// </summary>
+        FitNone,
```
jeremyVignelles, line 3625: Wouldn't it be better to explicitely set the enum values ?

mfkl: We could. Pinged upstream in case they want to add it.

mfkl: https://code.videolan.org/videolan/vlc/-/merge_requests/5295

PR #342, opened 2024-05-04
Avalonia VideoView fix

taivasverkko:
### Description of Change ###

In `VideoView` `MediaPlayer` handle was not properly assigned when using it in windows child component (eg. user control).
This caused MediaPlayer to popout to separate window. 

Added Initialized event handler that calls `Attach()` when control is initialized. 
Removed unnecessary Attach call from  `CreateNativeControlCore`. 

Updated **Avalonia.Sample** so that VideoView is displayed through user control and added play and stop buttons.

### Issues Resolved ### 
- fixes #525

### API Changes ###
 None

### Platforms Affected ### 
- Avalonia

- Avalonia VideoView works properly in child controls

None

Not applicable

### Testing Procedure ###
- Tested using Avalonia sample with windows desktop environment
 
### PR Checklist ###

### DIFF
--- a/samples/LibVLCSharp.Avalonia.Sample/ViewModels/MainWindowViewModel.cs
+++ b/samples/LibVLCSharp.Avalonia.Sample/ViewModels/MainWindowViewModel.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using Avalonia.Controls;
 using LibVLCSharp.Shared;
 
 namespace LibVLCSharp.Avalonia.Sample.ViewModels
@@ -6,6 +7,9 @@ namespace LibVLCSharp.Avalonia.Sample.ViewModels
     public class MainWindowViewModel : ViewModelBase, IDisposable
     {
         private readonly LibVLC _libVlc = new LibVLC();
+        
+        public MediaPlayer MediaPlayer { get; }
+        
         public MainWindowViewModel()
         {
             MediaPlayer = new MediaPlayer(_libVlc);
@@ -13,12 +17,20 @@ namespace LibVLCSharp.Avalonia.Sample.ViewModels
 
         public void Play()
         {
+            if (Design.IsDesignMode)
+            {
+                return;
+            }
+            
             using var media = new Media(_libVlc, new Uri("http://commondatastorage.googleapis.com/gtv-videos-bucket/sample/BigBuckBunny.mp4"));
             MediaPlayer.Play(media);
         }
-
-        public MediaPlayer MediaPlayer { get; }
-
+        
+        public void Stop()
+        {            
+            MediaPlayer.Stop();
+        }
+   
         public void Dispose()
         {
             MediaPlayer?.Dispose();
--- a/samples/LibVLCSharp.Avalonia.Sample/Views/MainWindow.xaml
+++ b/samples/LibVLCSharp.Avalonia.Sample/Views/MainWindow.xaml
@@ -1,22 +1,11 @@
 <Window xmlns="https://github.com/avaloniaui"
         xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
-        xmlns:vm="clr-namespace:LibVLCSharp.Avalonia.Sample.ViewModels;assembly=LibVLCSharp.Avalonia.Sample"
         xmlns:d="http://schemas.microsoft.com/expression/blend/2008"
         xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
         xmlns:v="clr-namespace:LibVLCSharp.Avalonia.Sample.Views"
-        xmlns:vlc="clr-namespace:LibVLCSharp.Avalonia;assembly=LibVLCSharp.Avalonia"
         mc:Ignorable="d" d:DesignWidth="800" d:DesignHeight="450"
         x:Class="LibVLCSharp.Avalonia.Sample.Views.MainWindow"
         Icon="/Assets/avalonia-logo.ico"
-        Title="LibVLCSharp.Avalonia.Sample"
-        Opened="OnOpened">
-
-    <Design.DataContext>
-        <vm:MainWindowViewModel />
-    </Design.DataContext>
-
-    <Panel>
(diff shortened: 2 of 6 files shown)


lawchingman (approved):

samples/LibVLCSharp.Avalonia.Sample/ViewModels/MainWindowViewModel.cs
```diff
@@ -1,24 +1,42 @@
...
             using var media = new Media(_libVlc, new Uri("http://commondatastorage.googleapis.com/gtv-videos-bucket/sample/BigBuckBunny.mp4"));
             MediaPlayer.Play(media);
         }
-
-        public MediaPlayer MediaPlayer { get; }
-
+        
+        public void Stop()
+        {
+            if (!MediaPlayer.IsPlaying)
```
mfkl, line 32: you can remove that `if` I think, since paused != stopped and I try to keep the sample minimal. Just redirect the call to the mediaplayer directly :)

mfkl:
Thank you for your contribution! Will test and merge soon.

samples/LibVLCSharp.Avalonia.Sample/ViewModels/MainWindowViewModel.cs
taivasverkko: Removed

mfkl:
Ran your code and all seems to run as designed.

samples/LibVLCSharp.Avalonia.Sample/Views/MainWindow.xaml
```diff
@@ -1,22 +1,11 @@
...
-        xmlns:vm="clr-namespace:LibVLCSharp.Avalonia.Sample.ViewModels;assembly=LibVLCSharp.Avalonia.Sample"
         xmlns:d="http://schemas.microsoft.com/expression/blend/2008"
         xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
         xmlns:v="clr-namespace:LibVLCSharp.Avalonia.Sample.Views"
-        xmlns:vlc="clr-namespace:LibVLCSharp.Avalonia;assembly=LibVLCSharp.Avalonia"
         mc:Ignorable="d" d:DesignWidth="800" d:DesignHeight="450"
         x:Class="LibVLCSharp.Avalonia.Sample.Views.MainWindow"
         Icon="/Assets/avalonia-logo.ico"
-        Title="LibVLCSharp.Avalonia.Sample"
-        Opened="OnOpened">
```
jeremyVignelles, line 12: This method should be removed in the .cs file as well

samples/LibVLCSharp.Avalonia.Sample/ViewModels/MainWindowViewModel.cs
```diff
@@ -1,24 +1,37 @@
...
     public class MainWindowViewModel : ViewModelBase, IDisposable
     {
         private readonly LibVLC _libVlc = new LibVLC();
+        
+        public MediaPlayer MediaPlayer { get; }
+        
         public MainWindowViewModel()
         {
             MediaPlayer = new MediaPlayer(_libVlc);
+            Play();
```
jeremyVignelles, line 16: I'd avoid putting this kind of logic in a constructor (it does not initialize the properties of the object). Isn't there a "Loaded" event that you can leverage on the Control to trigger the Play method ?

taivasverkko: Maybe we could utilize **Initialized** event in this case:

> Occurs when the styled element has finished initialization.

> The Initialized event indicates that all property values on the styled element have been set. When loading the styled element from markup, it occurs when ISupportInitialize.EndInit is called *and* the styled element is attached to a rooted logical tree. When the styled element is created by code and ISupportInitialize is not used, it is called when the styled element is attached to the visual tree.

https://reference.avaloniaui.net/api/Avalonia/StyledElement/D1745B44

jeremyVignelles: yes, please :)

samples/LibVLCSharp.Avalonia.Sample/Views/MainWindow.xaml
taivasverkko: Removed from OnOpened method from cs file along with unnecessary using statements.

samples/LibVLCSharp.Avalonia.Sample/ViewModels/MainWindowViewModel.cs
taivasverkko: Turns out that the Initialized event is fired before **Datacontext** is assigned to the control, so I think we could go with **DataContextChanged** event.  Video starts playing when viewmodel is assigned to usercontrol.

taivasverkko: Pushed changes.

lawchingman (approved):

lawchingman (approved):

mfkl:
Is `axaml` and `xaml` interchangeable in Avalonia? If not, why use `axaml` instead of `xaml`?

taivasverkko:
> Is `axaml` and `xaml` interchangeable in Avalonia? If not, why use `axaml` instead of `xaml`?

From official documentation:
> AXAML File Extension
The file extension for XAML files used elsewhere is .xaml but due to technical issues integrating with Visual Studio, Avalonia UI uses its own .axaml extension - 'Avalonia XAML'.

https://docs.avaloniaui.net/docs/basics/user-interface/introduction-to-xaml

mfkl (approved):

mfkl:
Thanks for your first contribution @taivasverkko! Will try to publish to nuget later this week.

zikaura:
Has this been published to nuget yet? I'm on version 3.8.5 yet I'm still facing this issue :(

mfkl:
Yes this is published on nuget already. Open an issue.

PR #294, opened 2022-04-05
Unofficial release of LibVLCSharp.Avalonia to solve some issues

radiolondra:
### Description of Change ###

Created new VideoView.cs

### Issues Resolved ### 

- Embed LibVLCSharp Avalonia VLC VideoView in Avalonia UserControl

- Put and use an Avalonia UserControl on top of VideoView (VLC MediaPlayer), for example, a set of buttons/controls to control the VLC MediaPlayer.

- fixes #525
- fixes #408

### API Changes ###

 None

### Platforms Affected ### 
- Avalonia

### Behavioral/Visual Changes ###
- Users can embed VideoView in Avalonia UserControl (not only in a window)
- Users can put a UserControl on top of MediaPlayer

### Before/After Screenshots ### 
See (after)screenshots in LibVLCSharp.Avalonia.Unofficial.Samples README.md

### Testing Procedure ###
See the LibVLCSharp.Avalonia.Unofficial.Samples added to the repo.

### PR Checklist ###

- [ ] Rebased on top of the target branch at time of PR
- [ ] Changes adhere to coding standard

PR #344, opened 2024-05-07
Avalonia: Child content support for VideoView

taivasverkko:
### Description of Change ###

Allow content to be placed to **VideoView** so that it renders it on top of the media player. 

- Created bindable **Content** property for **VideoView**.
- When content is assigned or **VideoView** is attached to visual tree, tries to create a new transparent overlay window that is placed on top of the **VideoView**.
- Content is binded to windows content property.
- Uses **VideoView** Content and Bound property changed events to update overlay position.

Updated sample.

Solution is based on this PR: https://github.com/videolan/libvlcsharp/pull/294

### Issues Resolved ### 
- fixes #408

### API Changes ###
Added:
 - object VideoView.Content { get; set; } //Bindable Property

### Platforms Affected ### 

- Avalonia

If user places control inside VideoView it will be shown top of the video.

image

### Testing Procedure ###
- Tested with Windows 11

### PR Checklist ###

### DIFF
--- a/samples/LibVLCSharp.Avalonia.Sample/Views/VideoPlayer.axaml
+++ b/samples/LibVLCSharp.Avalonia.Sample/Views/VideoPlayer.axaml
@@ -17,11 +17,15 @@
 
         <vlc:VideoView Grid.Row="1" MediaPlayer="{Binding MediaPlayer}"
                        HorizontalAlignment="Stretch"
-                       VerticalAlignment="Stretch" />
-
-        <StackPanel Grid.Row="2" Orientation="Horizontal" Spacing="20" Margin="20" HorizontalAlignment="Center">
-            <Button Command="{Binding Play}">Play</Button>
-            <Button Command="{Binding Stop}">Stop</Button>
-        </StackPanel>
+                       VerticalAlignment="Stretch"
+                       PointerEntered="VideoViewOnPointerEntered"
+                       PointerExited="VideoViewOnPointerExited">
+            <Panel Name="ControlsPanel">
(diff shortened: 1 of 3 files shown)


jeremyVignelles:
Thanks for implementing this 🎉 

I'm not a fan of the Min/MaxContentOpacity properties, because it seems specific to your use case and, if I remember correctly, there is no such thing on the WPF airspace hack.

Otherwise, it's looking good (though I didn't test) and I'm looking forward having that fixed in the Avalonia implementation :)

taivasverkko:
> Thanks for implementing this 🎉
> 
> I'm not a fan of the Min/MaxContentOpacity properties, because it seems specific to your use case and, if I remember correctly, there is no such thing on the WPF airspace hack. (that being said, if @mfkl is OK with this, we might as well add the feature to the WPF implementation)
> 
> Otherwise, it's looking good (though I didn't test) and I'm looking forward having that fixed in the Avalonia implementation :)

You are right. At first I thought there would be an use case for it but I've been tinkering with this by building an media player element for Avalonia and have come to the conclusion that its unnecessary. I will remove those properties. 

If someone comes along with actual use case for those we can implement them back later.

taivasverkko:
Converted this PR to draft. I was thinkin that it would be beneficial to get **PointerEntered**/**PointerExited** events from VideoView. Maybe other events aswell?

mfkl:
Tested, seems to work fine, thanks!

We'll likely have some fixes once it is used by a few people in production, this kind of thing is tricky to get right on the first go with all the edge cases (see the WPF history https://github.com/videolan/libvlcsharp/commits/3.x/src/LibVLCSharp.WPF), but that's alright.

mfkl:
Thanks for this! Will push to nuget tomorrow.

PR #352, opened 2024-06-18
Adding a MediaPlayerElement to Maui

jonx:
### Description of Change ###

This draft PR includes the implementation of a MediaPlayerElement for Maui. The changes include:
- Added new MediaPlayerElement control to LibVCLSharp.MAUI
- Added the sample LibVLCSharp.MAUI.Sample.MediaElement based on the Forms sample
- Added the sample LibVLCSharp.MAUI.Sample.MediaElement.Multiple that aims at showing how to style and use two instances of the control on the same page - but it's not yet doing that. Both controls are the same style currently

### Issues Resolved ### 

- fixes #637

### Other Changes ###

 - The SvgTintColorBehavior.cs was meant to change the color or the SVG icons but it's not working and can be ignored

### Platforms Affected ### 

- Core (all platforms)
- Android

None

Not applicable

### Testing Procedure ###

Not applicable

### PR Checklist ###


jonx:
The implementation is close to complete for Android but probably needs further testing.

1. I removed the dependance to AwesomeFont and replaced it with equivalent SVG files. The problem is that I can't change their color? I tried to implement a behavior to do that - it's in the project - but it's not working
2. I replaced Messaging center by WeakEventManager 
8. The checks in the track list popup should be more constrained. The controls there need some refinement.
9. The lock slider has some initial drawing issues. The icon only appears once you scroll once for some reason.
11. This is only an Android port, but I suspect getting the iOS version should be quite easy given all we have here is cross-platform.
12. The video starts playing before the player is visible. Maybe it's related to how the player is embeded.

Please review and provide feedback.

mfkl:
image

Looking good so far :-)

mfkl:
I'm wondering whether it would be possible to keep the same code for XF and MAUI and just use #ifdef? Depends how much code is the same or different, besides namespaces.
Also I don't know when we will drop XF support, so going through the effort of supporting both from a single code base might not be worth it. Not sure.

Will attempt to run the iOS sample next week, keep you posted.

jonx:
I guess it's possible in theory but I'm not sure how you're going to manage this. The way I see this is that it's probably less work to maintain both projects rather than having to add all the #ifdefs. Maybe there is an easy way to do such thing that I don't know of?

mfkl:
After some setup and deployment struggles, I managed to run the sample on an iOS device.

image

Buttons are present and working, just invisible.

mfkl:
Managed to display the play/pause icons in iOS with a hack. https://github.com/mfkl/libvlcsharp/commit/66d3d528a795141391124c06e320c09acddc4012;

Still need to figure out how to include them transitively (from libvlcsharp.maui) as they were not present in the app bundle.

MAUI expects png references from XAML or C# as it translates the svgs.

The path from the generic theme appears to be from the app bundle, not the code project.

src/LibVLCSharp.MAUI/LibVLCSharp.MAUI.csproj
```diff
@@ -13,13 +13,121 @@
...
+	</PropertyGroup>
 
-	<ItemGroup>
+	<PropertyGroup Condition="'$(Configuration)|$(TargetFramework)|$(Platform)'=='Release|net7.0-android|AnyCPU'">
+	  <TreatWarningsAsErrors>False</TreatWarningsAsErrors>
+	</PropertyGroup>
+
+	<PropertyGroup Condition="'$(Configuration)|$(TargetFramework)|$(Platform)'=='Release|net6.0-ios|AnyCPU'">
+	  <TreatWarningsAsErrors>False</TreatWarningsAsErrors>
+	</PropertyGroup>
```
mfkl, line 33: I think these can be removed.

src/LibVLCSharp.MAUI/LibVLCSharp.MAUI.csproj
```diff
@@ -13,13 +13,121 @@
...
+  </ItemGroup>
+
+  <ItemGroup>
+    <AndroidResource Remove="Handlers\**" />
+    <Compile Remove="Handlers\**" />
+    <EmbeddedResource Remove="Handlers\**" />
+    <MauiCss Remove="Handlers\**" />
+    <MauiXaml Remove="Handlers\**" />
+    <None Remove="Handlers\**" />
+  </ItemGroup>
```
mfkl, line 46: these lines can be removed too I guess

mfkl:
I'm thinking we can remove the Shared folder and namespace.

mfkl:
For the image issue on iOS, Maui is supposed to translate them into pngs during build with proper size options but there are various bugs preventing it from doing it transitively. With a BundleResource I managed to have them in the app bundle but they're not translated into pngs and iOS cannot read SVG natively...

jonx:
Alright, thanks for the review. I'll update the PR based on your comments and will revert back the Fontawesome removal.

TheSundayDev:
hi, 
can I ask if you have an estimation in order to be live? thans a lot!

TheSundayDev:
Hi,
MAUI ios and android for now.
thanks

jonx:
Alright, in this case, if you're in a hurry, my PR code is working already and not much will change before the final version on Android. We're only having this bug on iOS that is not including the button icons from the library to the main project. I'm just going to revert to using fonts like the original xamarin was doing. Meaning you can potentially start integrating it in your project. If not, stay tuned for my update next week. Then it will probably take a little longer for the VLC# people to accept the PR.

mfkl:
> hi, can I ask if you have an estimation in order to be live? thans a lot!

Hi, we do not provide delivery estimate for opensource work, it is released when it's ready. If you'd like to speed up the process, you may check out the PR's branch locally and help improve it by fixing the remaining issues. Thanks.

jonx:
I'm closing this PR to sync my fork and create a new PR.

PR #370, opened 2024-11-14
Maui winui

mfkl:
Getting lots of these
```
DEP0900: Failed to unregister application "49861f9e-48d3-45c0-b6a8-f591ad6d2010_1.0.0.1_x64__9zz4h110yvjzm". [0x80073CFA] 
```
Unsure why. Impossible to uninstall existing appx that match this app ID.

On 3.x, on WinUI, the `LibVLC` object is built using string parameters retrieved from an initialized VideoView (see the plan WinUI sample code). Will have to re-work the code to handle this in the MAUI sample for WinUI support.

### DIFF
--- a/samples/LibVLCSharp.MAUI.Sample/AppShell.xaml
+++ b/samples/LibVLCSharp.MAUI.Sample/AppShell.xaml
@@ -5,10 +5,5 @@
     xmlns:x="http://schemas.microsoft.com/winfx/2009/xaml"
     xmlns:local="clr-namespace:LibVLCSharp.MAUI.Sample"
     Shell.FlyoutBehavior="Disabled">
-
-    <ShellContent
-        Title="Home"
-        ContentTemplate="{DataTemplate local:MainPage}"
-        Route="MainPage" />
-
+    <ShellContent ContentTemplate="{DataTemplate local:MainPage}" Route="MainPage" />
 </Shell>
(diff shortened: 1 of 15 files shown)


jonx:
Have you tried to unregister the app manually?
In a powershell: Get-AppxPackage *49861f9e-48d3-45c0-b6a8-f591ad6d2010*
Then Remove-AppxPackage -Package full package name, in my case the full name is:  49861f9e-48d3-45c0-b6a8-f591ad6d2010_1.0.0.1_x64__9zz4h110yvjzm

mfkl:
Progress. Need to check that https://github.com/videolan/libvlcsharp/pull/370/commits/267bf18a11f59600a72c36c51d19d3c43142ce27 doesn't break other UI frameworks.

mfkl:
Without https://github.com/videolan/libvlcsharp/commit/267bf18a11f59600a72c36c51d19d3c43142ce27, there are errors like 
```
Exception thrown at 0x00007FFA3639B699 (KernelBase.dll) in LibVLCSharp.MAUI.Sample.exe: WinRT originate error - 0x800F1000 : 'Cannot apply a Style with TargetType 'LibVLCSharp.Platforms.Windows.VideoViewBase' to an object of type 'Microsoft.UI.Xaml.Controls.Control'.'.
Microsoft.ui.xaml.dll!00007FF99E8243D2: 800F1000 - E_NER_INVALID_OPERATION
```

With it, though, our windows video view initialization logic is broken since `OnApplyTemplate` never gets called. Reworking this view hierarchy would likely be a breaking change, for Uno, UWP and WinUI.

mfkl:
There is some exception when shutting down the Media Element sample, very little info with debugger attached..

mfkl:
Should be good enough for an initial release.

mfkl:
Potential breaking change in `src/LibVLCSharp/Themes/Generic.xaml` but no way around it, due to a MAUI/WinUI bug. Thoughts? @jeremyVignelles

jeremyVignelles (approved):
I think the breaking change should be quite rare given that we didn't document that style. In any case, the fix should be straightforward.

PR #373, opened 2025-01-15
Maui winui master

mfkl:


jasells:
Nice! I was just about to work on this myself!

There is one more target that would be good to add: pure .Net 8 core.  Adding this targetFramework would allow the lib to be added to XForms-style projects that don't use the shared-project style.  I ran into this recently trying to add the libvlc-sharp.Maui package to a project I recently ported to Maui from XForms, and still has x-platform projects and platform-projects.  

The Maui-UI code is in a `<TargetFramework>net8.0</TargetFramework>`  project, and thus this package currently is not compatible.

jasells:
See this [PR](https://github.com/videolan/libvlcsharp/pull/374).  

I assume github is not the main repo, since this PR appears to have already been completed?

mfkl:
Hi, thanks for your comments, will have a look at your PR shortly.

> added to XForms-style projects that don't use the shared-project style

Didn't know that was supported by maui. Not sure how many projects will use that structure for maui projects but adding net8.0 shouldn't be a problem.

> I assume github is not the main repo, since this PR appears to have already been completed?

Main repo is indeed https://code.videolan.org/videolan/LibVLCSharp, but the github one is a mirror with the CI that runs on azure devops, because we (vlc) don't have windows runners and I need those :). PRs should be opened on GitHub to this repo 👍

jasells:
> Didn't know that was supported by maui. Not sure how many projects will use that structure for maui projects but adding net8.0 shouldn't be a problem.

Technically, it is not official, but this [template pack](https://github.com/[mattleibow/MultiHeadMauiTemplates](https://github.com/mattleibow/MultiHeadMauiTemplates)) makes it easier.  I think it is useful for porting XForms projects -> Maui.  Much easier than manually moving all the source files into new project structure.   And, the Maui [packages ](https://www.nuget.org/packages/Microsoft.Maui.Controls/8.0.100#dependencies-body-tab) support .Netstandard2.0,2.1, and .Net8/9, so definitely supported, even if not using this template, can still put UI it's own .Net8/9/standard x-platform project.

This has been the hardest part of porting things since most OSS Maui libs don't consider this scenario.  Wish more did and also targeted .Netstandard, since that would make porting even easier.   Since libvlcsharp has .NetStandard targets, libvlcSharp.Maui _should_ be able to also target .NetStandard 2.0/1? I just thought about that.

But I also understand why MS is only showing the shared-project style in examples to reduce the complexity of needing to understand DI and moving it to compile-time DI (a.k.a. multi-targeting).

mfkl:
Understood, thanks for the details. I don't work on Xamarin/MAUI consumer apps anymore so did not know that :)

> Since libvlcsharp has .NetStandard targets, libvlcSharp.Maui should be able to also target .NetStandard 2.0/1? I just thought about that.

Yep, would that be useful to you? Not sure if that could cause tooling issues for consumers but probably not.

jasells:
> > Since libvlcsharp has .NetStandard targets, libvlcSharp.Maui should be able to also target .NetStandard 2.0/1? I just thought about that.
> 
> Yep, would that be useful to you? Not sure if that could cause tooling issues for consumers but probably not.

In short, it wouldn't be extremely useful to me, personally _at this point_.  However, this scenario has been the biggest hurdle porting Xforms ->Maui, and probably would help many devs porting Xform -> maui in the future if more lib code would also target netstandard2/2.1.    So far, it seems very few 3rd party (non-MS) Maui libs target .NetStandard, only .Net6/7/8.   I guess few dev understand/appreciate the advantage of .Netstandard for libs, or want to take the time to support it?  I dunno.


As far as libVlcSharp.Maui goes, it _might_ require a little more refactor?  Maybe not, maybe just another `#if netstandard`  block that is the same as the .net8 core target?  As far as consumers, no, it wouldn't affect them, other than making it easier for those porting .Netstandard UI libs from XForms and allowing Maui libs to be defined as .Netstandard OR .Net8.

I'll open another PR to update that, the readme to include Windows, and include the platform-package refs. I forgot to do that, but it at least makes it easier to install if they show up in the transitive package list instead of having to manually search and install them all individually.  

I wish I knew how to make those native build targets install automatically without re-packaging them directly into the libVlcsharp.Maui package.  I played around with that for a while today.

PR #384, opened 2025-03-25
libvlc: add helper delay function

mfkl:


jeremyVignelles (approved):

mfkl:
After talking with core devs, I'm not sure it makes sense to add this and whether it is even possible to use it properly within the audio callbacks..

PR #387, opened 2025-05-06
Lib vlc options builder

jdswcz:
### Description of Change ###

Added a builder pattern for `LibVLCOptions`.

This change introduces a `LibVLCOptionsBuilder` class to simplify and structure the creation of LibVLC initialization options. The builder provides a fluent, discoverable API for composing commonly used options, improving readability and reducing the chance of misconfigured strings.

Maintains compatibility with old LibVLC constructors to avoid problems for developers when updating to version with this change.

### Issues Resolved ### 

Fixes #654 (Parameter `--audio-resampler` for UWP causes audio to stutter)

### API Changes ###

Added:
 - class `LibVLCOptionsBuilder`
 - LibVLCOptionsBuilder AddOption(string option)
 - LibVLCOptionsBuilder AddOptions(IEnumerable<string> options)
 - LibVLCOptionsBuilder SetVerbosity(int level)
 - LibVLCOptionsBuilder EnableDebugLogs()
 - LibVLCOptionsBuilder SetAudioResampler(string resampler)
 - LibVLCOptionsBuilder SetAudioOutput(string output)
 - LibVLCOptionsBuilder DisableVideo()
 - LibVLCOptionsBuilder DisableAudio()
 - LibVLCOptionsBuilder SetNetworkCaching(int milliseconds)
 - LibVLCOptionsBuilder SetFileCaching(int milliseconds)
 - LibVLCOptionsBuilder EnableHardwareDecoding(string method)
 - LibVLCOptionsBuilder DropLateFrames()
 - LibVLCOptionsBuilder SkipFrames()
 - LibVLCOptionsBuilder UseXlib(bool enabled)
 - LibVLCOptionsBuilder Build() : LibVLCOptions

#if UWP || ANDROID
 - LibVLCOptionsBuilder DisableDefaultResampler()
#endif

 - class `LibVLCOptions`
 - LibVLCOptions(string[] options)
 - string[] LibVLCOptions.Options { get; }
 - static LibVLCOptions LibVLCOptions.Empty { get; }

Removed.

- LibVLC.PatchOptions()

### Platforms Affected ### 

- Core (all platforms)
- iOS
- Android
- UWP

None

Not applicable

### Testing Procedure ###

- Verified that `LibVLCOptionsBuilder` produces correct and expected LibVLC option strings
- Confirmed that usage of the builder allows to avoid problematic parameter `--audio-resampler=speex_resampler"` on UWP

### PR Checklist ###

- [x] New API is documented and tested

### DIFF
--- a/src/LibVLCSharp/Shared/LibVLC.cs
+++ b/src/LibVLCSharp/Shared/LibVLC.cs
@@ -206,7 +206,7 @@ namespace LibVLCSharp.Shared
         /// <code>
         /// var options = new LibVLCOptionsBuilder()
         ///     .EnableDebugLogs()
-        ///     .SetAudioResampler(\"soxr\")
+        ///     .SetAudioResampler("soxr")
         ///     .Build();
         ///
         /// using var libvlc = new LibVLC(options);
@@ -216,11 +216,10 @@ namespace LibVLCSharp.Shared
         /// <param name="options">An instance of <see cref="LibVLCOptions"/> containing libvlc configuration parameters.</param>
 
         public LibVLC(LibVLCOptions options)
-    : base(() => MarshalUtils.CreateWithOptions(options.Options, Native.LibVLCNew), Native.LibVLCRelease)
+        : base(() => MarshalUtils.CreateWithOptions(options.Options, Native.LibVLCNew), Native.LibVLCRelease)
         {
             _gcHandle = GCHandle.Alloc(this);
         }
-
 
         /// <summary>
         /// Create and initialize a libvlc instance.
--- a/src/LibVLCSharp/Shared/LibVLC.cs
+++ b/src/LibVLCSharp/Shared/LibVLC.cs
@@ -190,6 +190,39 @@ namespace LibVLCSharp.Shared
         }
 
         /// <summary>
+        /// Create and initialize a libvlc instance using a <see cref="LibVLCOptions"/> object.
+        /// This constructor allows for strongly-typed, flexible configuration of libvlc command-line parameters,
+        /// including platform-specific defaults, logging verbosity, resamplers, caching, and rendering settings.
+        /// LibVLC may create threads. Therefore, any thread-unsafe process
+        /// initialization must be performed before calling libvlc_new(). In particular:
+        /// <para>- setlocale() and textdomain(),</para>
+        /// <para>- setenv(), unsetenv(), and putenv(),</para>
+        /// <para>- On Microsoft Windows, SetErrorMode().</para>
+        /// On POSIX systems, certain signals must be handled properly (e.g., SIGCHLD must not be ignored).
+        /// <para/> This will throw a <see cref="VLCException"/> if the native libvlc libraries cannot be found or loaded.
+        /// <para/> It may also throw a <see cref="VLCException"/> if the LibVLC and LibVLCSharp major versions do not match.
+        /// See https://code.videolan.org/videolan/LibVLCSharp/-/blob/master/docs/versioning.md for more info about the versioning strategy.
+        /// <example>
+        /// <code>
+        /// var options = new LibVLCOptionsBuilder()
+        ///     .EnableDebugLogs()
+        ///     .SetAudioResampler(\"soxr\")
+        ///     .Build();
+        ///
+        /// using var libvlc = new LibVLC(options);
+        /// </code>
+        /// </example>
+        /// </summary>
+        /// <param name="options">An instance of <see cref="LibVLCOptions"/> containing libvlc configuration parameters.</param>
+
+        public LibVLC(LibVLCOptions options)
+    : base(() => MarshalUtils.CreateWithOptions(options.Options, Native.LibVLCNew), Native.LibVLCRelease)
+        {
+            _gcHandle = GCHandle.Alloc(this);
+        }
+
+
+        /// <summary>
         /// Create and initialize a libvlc instance.
         /// This functions accept a list of &quot;command line&quot; arguments similar to the
         /// main(). These arguments affect the LibVLC instance default configuration.
@@ -239,10 +272,8 @@ namespace LibVLCSharp.Shared
         /// <param name="options">list of arguments, in the form "--option=value"</param>
         /// <returns>the libvlc instance or NULL in case of error</returns>
         public LibVLC(params string[] options)
-            : base(() => MarshalUtils.CreateWithOptions(PatchOptions(options), Native.LibVLCNew), Native.LibVLCRelease)
-        {
-            _gcHandle = GCHandle.Alloc(this);
-        }
+            : this(new LibVLCOptionsBuilder().AddOptions(options).Build()) { }
+
 
         /// <summary>
         /// Create and initialize a libvlc instance.
@@ -286,31 +317,11 @@ namespace LibVLCSharp.Shared
         /// <param name="enableDebugLogs">enable verbose debug logs</param>
         /// <param name="options">list of arguments (should be NULL)</param>
         public LibVLC(bool enableDebugLogs, params string[] options)
-            : base(() => MarshalUtils.CreateWithOptions(PatchOptions(options, enableDebugLogs), Native.LibVLCNew), Native.LibVLCRelease)
-        {
-            _gcHandle = GCHandle.Alloc(this);
-        }
-
-        /// <summary>
-        /// Make dirty hacks to include necessary defaults on some platforms.
-        /// </summary>
-        /// <param name="options">The options given by the user</param>
-        /// <param name="enableDebugLogs">enable debug logs</param>
-        /// <returns>The patched options</returns>
-        static string[] PatchOptions(string[] options, bool enableDebugLogs = false)
-        {
-            if(enableDebugLogs)
-            {
-                options = options.Concat(new[] { "--verbose=2" }).ToArray();
-            }
-#if UWP
-            return options.Concat(new[] {"--aout=winstore", "--audio-resampler=speex_resampler"}).ToArray();
-#elif ANDROID
-            return options.Concat(new[] {"--audio-resampler=soxr"}).ToArray();
-#else
-            return options;
-#endif
-        }
+            : this(new LibVLCOptionsBuilder()
+                        .AddOptions(options)
+                        .EnableDebugLogs()
+                        .Build())
+        { }
 
         /// <summary>
         /// Dispose of this libvlc instance
--- a/src/LibVLCSharp/Shared/LibVLCOptions.cs
+++ b/src/LibVLCSharp/Shared/LibVLCOptions.cs
@@ -210,10 +210,5 @@ namespace LibVLCSharp.Shared
         {
             Options = options?.ToArray() ?? new string[0];
         }
-
-        /// <summary>
-        /// Gets an empty LibVLCOptions instance.
-        /// </summary>
-        public static LibVLCOptions Empty => new LibVLCOptions(new string[0]);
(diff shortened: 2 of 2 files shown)


mfkl:
Thanks for the contribution!

I'm gonna need some time to review and think if this approach is acceptable, so bear with me please. From an initial glimpse, a few things come to mind:
- https://github.com/videolan/libvlcsharp/blob/3.x/src/LibVLCSharp/Shared/MediaConfiguration.cs some libvlc options can also be configured through the media object. But not all options work through the Media, some only work through the libvlc constructor. The list of options that work with each approach is undocumented.
- libvlc options are officially unsupported, even though many consumers use them. What this means is that there is no guarantee of any API compatibility between major versions. Introducing C# functions as part of the rest of LibVLCSharp (which does follow API best practice, as LibVLC functions do as well) will definitely confuse users. Warnings and docs is necessary, but may not be sufficient.

Let me think about this.. thanks.

jdswcz:
> Thanks for the contribution!
> 
> I'm gonna need some time to review and think if this approach is acceptable, so bear with me please. From an initial glimpse, a few things come to mind:
> 
> * https://github.com/videolan/libvlcsharp/blob/3.x/src/LibVLCSharp/Shared/MediaConfiguration.cs some libvlc options can also be configured through the media object. But not all options work through the Media, some only work through the libvlc constructor. The list of options that work with each approach is undocumented.
> * libvlc options are officially unsupported, even though many consumers use them. What this means is that there is no guarantee of any API compatibility between major versions. Introducing C# functions as part of the rest of LibVLCSharp (which does follow API best practice, as LibVLC functions do as well) will definitely confuse users. Warnings and docs is necessary, but may not be sufficient.
> 
> Let me think about this.. thanks.

No worries and take your time. Unfortuantely, I didn't know all the details you mentioned and this is my first PR for LibVLCSharp. Personally, I use libvlc options as well, for example, to configure subtitles design (if libvlc is used to render subtitles).

To clarify my approach: There were originally two LibVLC.cs constructors that already allowed passing any arguments to the libvlc through LibVLCSharp. These were later patched by PatchOptions function while its behavior caused the bug we already discussed. Generally, you are forcing some libvlc options without providing to devs using LibVLCSharp an ability to control the options passed to libvlc. Of course, I could just add some simple parameter to constructor to avoid this behavior. However, I wanted to do it thoroughly. Thus I added builder for libvlc options with support of some basic options as well as adding any generic option user may like. Of course, I kept the compatibility with previous constructors, so updating to new version (in case this PR is approved), won't break current integrations.

Please, consider my PR as concept that should be extended according to https://wiki.videolan.org/VLC_command-line_help/. While I understand (but didn't know in advance) that it is officially unsupported, I believe that using my approach and providing easy to use C# code to build libvlc options will benefit most users and make libvlcsharp integrations easier for everyone.

If you wish, I am willing to help to extend the options builder and document it according to what is (or is not) supported throughout different libvlc versions. I would just need correct information to be able to complete this job.

src/LibVLCSharp/Shared/LibVLC.cs
```diff
@@ -189,6 +189,39 @@ internal struct Native
...
+        /// <para>- On Microsoft Windows, SetErrorMode().</para>
+        /// On POSIX systems, certain signals must be handled properly (e.g., SIGCHLD must not be ignored).
+        /// <para/> This will throw a <see cref="VLCException"/> if the native libvlc libraries cannot be found or loaded.
+        /// <para/> It may also throw a <see cref="VLCException"/> if the LibVLC and LibVLCSharp major versions do not match.
+        /// See https://code.videolan.org/videolan/LibVLCSharp/-/blob/master/docs/versioning.md for more info about the versioning strategy.
+        /// <example>
+        /// <code>
+        /// var options = new LibVLCOptionsBuilder()
+        ///     .EnableDebugLogs()
+        ///     .SetAudioResampler(\"soxr\")
```
mfkl, line 209: why the `\` ? Don't think that's needed, you can remove them.

src/LibVLCSharp/Shared/LibVLC.cs
```diff
@@ -189,6 +189,39 @@ internal struct Native
...
+        ///     .Build();
+        ///
+        /// using var libvlc = new LibVLC(options);
+        /// </code>
+        /// </example>
+        /// </summary>
+        /// <param name="options">An instance of <see cref="LibVLCOptions"/> containing libvlc configuration parameters.</param>
+
+        public LibVLC(LibVLCOptions options)
+    : base(() => MarshalUtils.CreateWithOptions(options.Options, Native.LibVLCNew), Native.LibVLCRelease)
```
mfkl, line 219: missing indentation

src/LibVLCSharp/Shared/LibVLCOptions.cs
```diff
@@ -0,0 +1,217 @@
...
+        /// <param name="options">The options to include.</param>
+        public LibVLCOptions(IEnumerable<string>? options)
+        {
+            Options = options?.ToArray() ?? new string[0];
+        }
+
+        /// <summary>
+        /// Gets an empty LibVLCOptions instance.
+        /// </summary>
+        public static LibVLCOptions Empty => new LibVLCOptions(new string[0]);
```
mfkl, line 215: this isn't called anywhere, can remove.

src/LibVLCSharp/Shared/LibVLCOptions.cs
```diff
@@ -0,0 +1,217 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace LibVLCSharp.Shared
+{
+    /// <summary>
+    /// Fluent builder for constructing LibVLC options.
+    /// Mirrors many common libvlc command line parameters.
```
mfkl, line 9: should add a link to https://wiki.videolan.org/VLC_command-line_help in this comment/doc and mention that libvlc options are provided as-is, without any guarantee of forward/backward compatibility.

src/LibVLCSharp/Shared/LibVLC.cs
jdswcz: You are right, it's not necessary in XML comment. Removed.

src/LibVLCSharp/Shared/LibVLC.cs
jdswcz: Fixed indent

src/LibVLCSharp/Shared/LibVLC.cs
```diff
@@ -189,6 +189,39 @@ internal struct Native
...
+        {
+            _gcHandle = GCHandle.Alloc(this);
+        }
+
+
```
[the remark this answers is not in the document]
jdswcz: Extra space removed

src/LibVLCSharp/Shared/LibVLCOptions.cs
jdswcz: While I understand your point I would recommend to keep this. If someone wants to instantiate LibVLC class without any parameters, he can just call LibVLCOptions.Empty instead of having to construct new empty options.

src/LibVLCSharp/Shared/LibVLCOptions.cs
jdswcz: Added as requested.

src/LibVLCSharp/Shared/LibVLCOptions.cs
mfkl: > If someone wants to instantiate LibVLC class without any parameters

Then they can just write `new LibVLC();`

jdswcz: Okay. Apologies for the delay. Removed as requested.

src/LibVLCSharp/Shared/LibVLC.cs
```diff
@@ -286,31 +316,11 @@ public LibVLC(params string[] options)
...
-            return options.Concat(new[] {"--aout=winstore", "--audio-resampler=speex_resampler"}).ToArray();
-#elif ANDROID
-            return options.Concat(new[] {"--audio-resampler=soxr"}).ToArray();
-#else
-            return options;
-#endif
-        }
+            : this(new LibVLCOptionsBuilder()
+                        .AddOptions(options)
+                        .EnableDebugLogs()
```
mfkl, line 321: I think the logs are enabled when running with `new LibVLC(enableDebugLogs: false)` here?

src/LibVLCSharp/Shared/LibVLCOptions.cs
```diff
@@ -0,0 +1,214 @@
...
+
+        /// <summary>
+        /// Builds the final LibVLCOptions object including any platform-specific defaults.
+        /// </summary>
+        public LibVLCOptions Build()
+        {
+            var finalOptions = _options;
+
+            if (_includeDebugLogs)
+                finalOptions.Add("--verbose=2");
```
mfkl, line 180: What happens when someone writes
```
.EnableDebugLogs().SetVerbosity(1);
```

Might be confusing. I guess it's one or the other? Maybe worth a comment.

mfkl (approved):

src/LibVLCSharp/Shared/LibVLCOptions.cs
```diff
@@ -0,0 +1,214 @@
...
+
+namespace LibVLCSharp.Shared
+{
+    /// <summary>
+    /// Fluent builder for constructing LibVLC options.
+    /// Mirrors many common libvlc command line parameters.
+    /// See https://wiki.videolan.org/VLC_command-line_help for more details about available parameters.
+    /// OPTIONS ARE PROVIDED AS-IS, WITHOUT ANY GUARANTEE OF FORWARD/BACKWARD COMPATIBILITY.
+    /// </summary>
+    public class LibVLCOptionsBuilder
```
jeremyVignelles, line 13: I'd move this class in its own file. I would always have a bad time trying to find that LibVLCOptionsBuilder.cs that doesn't exist...

src/LibVLCSharp/Shared/LibVLCOptions.cs
```diff
@@ -0,0 +1,214 @@
...
+        {
+            _options.Add("--skip-frames");
+            return this;
+        }
+
+        /// <summary>
+        /// Enables or disables use of Xlib.
+        /// </summary>
+        /// <param name="enabled">True to enable Xlib, false to disable.</param>
+        public LibVLCOptionsBuilder UseXlib(bool enabled)
```
jeremyVignelles, line 145: Only on linux ?

jeremyVignelles (commented):
Agree with @mfkl : Options are a difficult topic because they're not part of the official API, might change at any time, and are mostly OS and plugins specific.

When it comes to Builder patterns, I'm used to extensions methods (see asp.net core service configuration), because that Builder class can get really heavy. However, since methods are one-liners there, not sure it's worth the overhead...

Just thinking out loud, no strong opinion here.

I'm just slightly leading against this PR: Options are already poorly documented, and you'll have to find the equivalent builder method to what you're doing, while it's quite easy (and more memory efficient) to add options to a static array directly.

mfkl:
> I'm just slightly leading against this PR: Options are already poorly documented, and you'll have to find the equivalent builder method to what you're doing, while it's quite easy (and more memory efficient) to add options to a static array directly.

Yeah.. you might be right.

Maybe we should just provide a flag to ignore the options we are currently forcing in LVS (which was the initial point of this PR).

mfkl:
superseded by https://github.com/videolan/libvlcsharp/pull/390

PR #418, opened 2026-03-03
fix:multi-plane support

ashish-066:


ashish-066:
@mfkl can you look into this and share your opinion?

mfkl:
it does not build

mfkl:
Multi plane access is actually already possible through the use of `Unsafe.Add` but we can add your approach as a clearer/convenience method. You need to update `VideoFormatCallback` to handle both callbacks path and then it should be fine. you could name the callback `LibVLCVideoMultiPlaneFormatCb`.

mfkl (changes requested):
approach works, few changes to make before merging. If you want to add a test or sample showing its use, much welcome

ashish-066:
@mfkl can you review the changes

mfkl:
did you use an AI for this?

src/LibVLCSharp/Shared/MediaPlayer.cs
```diff
@@ -2008,13 +2045,19 @@ private static void AudioCleanupCallback(IntPtr opaque)
...
+            ref uint height, IntPtr pitches, IntPtr lines);
+
         /// <summary>Callback prototype to configure picture buffers format.</summary>
         /// <param name="opaque">
         /// <para>private pointer as passed to libvlc_video_set_callbacks()</para>
         /// <para>(and possibly modified by</para>
         /// </param>
         [UnmanagedFunctionPointer(CallingConvention.Cdecl)]
-        public delegate void LibVLCVideoCleanupCb(ref IntPtr opaque);
+        public delegate void LibVLCVideoCleanupCb(IntPtr opaque);
```
mfkl: correct fix, but API break

mfkl:
did you actually test the interop works? In a sample or unit test?

ashish-066:
> did you use an AI for this?

no

src/LibVLCSharp/Shared/MediaPlayer.cs
ashish-066: > correct fix, but API break

You are right I made this change is that the native libVLC signature for the cleanup callback is void (*cleanup)(void *opaque). According to the libVLC headers, the parameter is passed as void *opaque, which maps to IntPtr in .NET rather than ref IntPtr.

https://code.videolan.org/videolan/vlc/-/blob/master/include/vlc/libvlc_media_player.h

Using ref IntPtr would correspond to void**, which doesn't match the native API.

ashish-066: i will change it

mfkl:
did you actually test the interop works? In a sample or unit test?

mfkl:
feel free to reopen with updates

PR #421, opened 2026-04-06
Improved Avalonia Sample Hack

nunofaria007:
Improved the previous LibVLCSharp.Avalonia.Samples VideoPlayer hack and made it more reusable.

### Description of Change ###

-Transformed VideoPlayer into a more reusable component with it's own MediaPlayer DirectProperty that takes ownership of VideoView.MediaPlayerProperty (...but does taking ownership even matter!?)
-The .axml now defines a ControlTemplate where VideoView's MediaPlayer and Content are TemplateBinded
-Removed any reference to a ViewModel in the VideoPlayer
-Now it is possible to set the overlay directly on VideoPlayer instead of VideoView

```
<Window>
   <VideoPlayer {Binding MediaPlayer}>
       <Panel> <- we can use the top level container to handle pointer events
           ...some other controls
       </Panel>
   </VideoPlayer>
</Window>
```

Note: VideoPlayer pointer events don't trigger, but we can wire them like this if really needed:

VideoPlayer.axml:
```
 <vlc:VideoView Content="{TemplateBinding Content}"
                           PointerEntered="VideoView_PointerEvent"
                           ...any other event needed
                           MediaPlayer="{TemplateBinding MediaPlayer}"/>
```
                           
VideoPlayer.axml.cs:
```
private void VideoView_PointerEvent(object? sender, PointerEventArgs e)
{
    this.RaiseEvent(e);
}
```

This could also be integrated into the library:

-VideoPlayer would become the new VideoView
-The old VideoView would become NativeVideoView

### Issues Resolved ### 
None

### API Changes ###
None

### Platforms Affected ### 

None

None

Not applicable

### PR Checklist ###

### DIFF
--- a/samples/LibVLCSharp.Avalonia.Sample/ViewModels/MainWindowViewModel.cs
+++ b/samples/LibVLCSharp.Avalonia.Sample/ViewModels/MainWindowViewModel.cs
@@ -22,7 +22,7 @@ namespace LibVLCSharp.Avalonia.Sample.ViewModels
                 return;
             }
             
-            using var media = new Media(_libVlc, new Uri("http://commondatastorage.googleapis.com/gtv-videos-bucket/sample/BigBuckBunny.mp4"));
+            using var media = new Media(_libVlc, new Uri("https://www.w3schools.com/tags/mov_bbb.mp4"));
             MediaPlayer.Play(media);
         }
         
--- a/samples/LibVLCSharp.Avalonia.Sample/Views/MainWindow.xaml
+++ b/samples/LibVLCSharp.Avalonia.Sample/Views/MainWindow.xaml
@@ -1,11 +1,26 @@
 <Window xmlns="https://github.com/avaloniaui"
-        xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
-        xmlns:d="http://schemas.microsoft.com/expression/blend/2008"
-        xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
-        xmlns:v="clr-namespace:LibVLCSharp.Avalonia.Sample.Views"
-        mc:Ignorable="d" d:DesignWidth="800" d:DesignHeight="450"
-        x:Class="LibVLCSharp.Avalonia.Sample.Views.MainWindow"
-        Icon="/Assets/avalonia-logo.ico"
-        Title="LibVLCSharp.Avalonia.Sample">
-    <v:VideoPlayer />
+		xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
+		xmlns:d="http://schemas.microsoft.com/expression/blend/2008"
+		xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
+		xmlns:v="clr-namespace:LibVLCSharp.Avalonia.Sample.Views"
+		mc:Ignorable="d" d:DesignWidth="800" d:DesignHeight="450"
+		x:Class="LibVLCSharp.Avalonia.Sample.Views.MainWindow"
+		Icon="/Assets/avalonia-logo.ico"
+		Title="LibVLCSharp.Avalonia.Sample">
+	<Grid Name="teste" RowDefinitions="Auto, *, Auto">
+		<Label Grid.Row="0" 
+			   Content="Video Player"
+			   HorizontalAlignment="Center"/>
+		<v:VideoPlayer Grid.Row="1" 
+					   MediaPlayer="{Binding MediaPlayer}"
+					   HorizontalAlignment="Stretch"
+					   VerticalAlignment="Stretch">
+			<Panel Name="ControlsPanel">
+				<StackPanel Orientation="Horizontal" VerticalAlignment="Bottom" Background="#900000FF" Spacing="20">
+					<Button Command="{Binding Play}" Margin="20">Play</Button>
+					<Button Command="{Binding Stop}" Margin="20">Stop</Button>
+				</StackPanel>
+			</Panel>
+		</v:VideoPlayer>
+	</Grid>
 </Window>
--- a/samples/LibVLCSharp.Avalonia.Sample/Views/MainWindow.xaml.cs
+++ b/samples/LibVLCSharp.Avalonia.Sample/Views/MainWindow.xaml.cs
@@ -1,4 +1,5 @@
 using Avalonia.Controls;
+using Avalonia.Input;
 using Avalonia.Markup.Xaml;
 
 namespace LibVLCSharp.Avalonia.Sample.Views
--- a/samples/LibVLCSharp.Avalonia.Sample/Views/VideoPlayer.axaml
+++ b/samples/LibVLCSharp.Avalonia.Sample/Views/VideoPlayer.axaml
@@ -1,31 +1,11 @@
-﻿<UserControl xmlns="https://github.com/avaloniaui"
-             xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
-             xmlns:d="http://schemas.microsoft.com/expression/blend/2008"
-             xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
-             xmlns:vlc="clr-namespace:LibVLCSharp.Avalonia;assembly=LibVLCSharp.Avalonia"
-             xmlns:viewModels="clr-namespace:LibVLCSharp.Avalonia.Sample.ViewModels"
-             mc:Ignorable="d" d:DesignWidth="800" d:DesignHeight="450"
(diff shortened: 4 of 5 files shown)


mfkl:
Hi,

Thanks for the PR.

I'm not sure about this change though. The idea of this sample is to _show_ a minimal implementation of the VideoView and keep it as simple and straightforward as possible. Real apps can build on this and add their own abstractions, but it's sort of out of scope for this repo. We have more complete samples here https://github.com//mfkl/libvlcsharp-samples if you'd like to contribute a more featured avalonia mediaplayer sample.

Additionally, your code removes the autoplay functionality and the controls overlay on hover.

Media URL indeed seems to be invalid, and it could be updated for all samples.

nunofaria007:
Hi, 

Originally i meant to submit the UserControl as part of the library and rework the sample too. Instead I took a simpler approach and changed the sample just to showcase the Control and let you decide if it would be worth to integrate it in the library. For the end-user it would mean having a straightforward Control without having to reproduce the hack seen in the sample just to keep the video from popping out of the window.

"Additionally, your code removes the autoplay functionality and the controls overlay on hover." - true, I was not sure what would be better, if letting the control trigger those events, or if letting the child control do it, so I ended up just commenting the code for reference. In my personal project I was handling the events through the child, but now on second thought others might want to access to those events without having to add a child, so it would make sense to wire them up. 

Other thing I noticed. When pressing F12 to use the DevTools we can't inspect the children of VideoView, only way around is calling AttachDevTools programatically then pressing F12. This will cause the extra Window to be listed. With the UserControl I provided this behavior changes. Pressing F12 immediately shows the child element in the MainWindow hierarchy. Calling AttachDevTools and pressing F12 will make the extra Window to be listed, but the hierarchy of MainWindow remains the same.

Old sample F12 - ControlsPanel doesn't show up
old sample F12

Old sample AttachDevTools + F12 - ControlsPanel shows up under Window
old sample AttachDevTools + F12

New sample F12 - ControlsPanel show up under VideoPlayer
new sample F12

New sample AttachDevTools + F12 - ControlsPanel show up under VideoPlayer and under Window
new sample AttachDevTools + F12

mfkl:
Ok I can see how having F12 could be helpful but I also try to limit unnecessary breaking changes. UserControl support was added in https://github.com/videolan/libvlcsharp/commit/8a5257b14136660c2f956520d958511f810a5e68 which could indeed be a nicer user experience if we provide it out of the box. Could this be done without breaking the existing VideoView API tho?

nunofaria007:
I was able to track down the reason why VideoView pops out if we just use it without creating a UserControl first.

When VideoView is inside a UserControl

inside UserControl

When VideoView is outside a UserControl (Pops out because _mediaPlayer.Hwnd = _platformHandle.Handle; is never reached)

outside UserControl

To solve this we need to call Attach() inside CreateNativeControlCore() and 

`Initialized += (_, _) => { Attach(); };
`

could be discarded if we update the if statement inside Attach() to 

` if (_mediaPlayer == null || _platformHandle == null)`

The result:

When VideoView is outside a UserControl

outside UserControl Attach on Create

When VideoView is inside a UserControl

inside UserControl Attach() on Create

This, in theory, shouldn't produce any breaking change.



As for the DevTools inspection, that can be solved by changing the contentChangedHandler 

```
contentChangedHandler = ContentProperty.Changed.AddClassHandler<VideoView>((s, content) =>
{
    LogicalChildren.Clear();

    if (content.NewValue is ILogical logical )
    {                    
        LogicalChildren.Add(logical);
    }

    s.UpdateOverlayPosition();
});
```

Now ControlsPanel is visible under VideoView
Logical Tree

However this has a side effect, the Parent will become VideoView instead of Window

ControlsPanel's Parent becomes VideoView when added to VideoView's LogicalChildren
ControlsPanel Parent

ControlsPanel's Parent is Window when not added to VideoView's LogicalChildren
ControlsPanel Parent no LogicChildren

To be honest, when someone writes 

```
<VideoView>
    <Panel Name="MyOverlay"/>
</VideoView>
```
they probably expect MyOverlay.Parent to return a reference to VideoView and not Window. But if right now there's someone out there that for some reason is doing something expecting the type Window, their code will break.

winkmichael:
I believe we are talking about the same issue??? https://github.com/winkmichael/EmbeddedVideoView.Avalonia

nunofaria007:
Not really. right now if we create a fresh Avalonia MVVM project, install LibVLCSharp 3.9.6 and write this UI:

image

This is what we get if we instantiate MediaPlayer inside the VM constructor:
image

**This is happening because every Attach() call is happening before CreateNativeControlCore() sets _platformHandle.** 

Timeline:

1. MediaPlayer's setter calls Attach() -> if(_mediaPlayer == null || _platformHandle == null || !IsInitialized) return; evaluates to TRUE and Attach() returns

    - _mediaPlayer is not null
    - _platformHandle is null
    - IsInitialized is FALSE

2. Initialized's lambda calls Attach() -> if(_mediaPlayer == null || _platformHandle == null || !IsInitialized) return; evaluates to TRUE and Attach() returns

    - _mediaPlayer is not null
    - _platformHandle is null
    - IsInitialized is TRUE
    
3. CreateNativeControlCore() sets _platformHandle -> after this nothing calls Attach()

    _mediaPlayer.Hwnd = _platformHandle.Handle; was never reached

**If we write a UserControl like VideoPlayer, the order changes**

Timeline:
1. MediaPlayer's setter calls Attach() -> if(_mediaPlayer == null || _platformHandle == null || !IsInitialized) return; evaluates to TRUE and Attach() returns

    - _mediaPlayer is not null
    - _platformHandle is null
    - IsInitialized is FALSE

2. CreateNativeControlCore() sets _platformHandle

3. Initialized's lambda calls Attach() -> if(_mediaPlayer == null || _platformHandle == null || !IsInitialized) return; evaluates to FALSE 

    - _mediaPlayer is not null
    - _platformHandle is not null
    - IsInitialized is TRUE
    
    _mediaPlayer.Hwnd = _platformHandle.Handle; is reached

**If we go back to the first example but instantiate MediaPlayer at a later time (let's say inside a Command and MediaPlayer is an ObservableProperty)**

Timeline:
1. Initialized's lambda calls Attach() -> if(_mediaPlayer == null || _platformHandle == null || !IsInitialized) return; evaluates to TRUE and Attach() returns

    - _mediaPlayer is is null
    - _platformHandle is null
    - IsInitialized is TRUE

2. CreateNativeControlCore() sets _platformHandle 

3. MediaPlayer's setter calls Attach() -> if(_mediaPlayer == null || _platformHandle == null || !IsInitialized) return; evaluates to FALSE

    - _mediaPlayer is not null
    - _platformHandle is not null
    - IsInitialized is TRUE

    _mediaPlayer.Hwnd = _platformHandle.Handle; is reached

nunofaria007:
To summarize everything so far:

When I started my personal project, I couldn't wrap my head around the fact that I couldn't write 

```
<Window>
  <VideoView MediaPlayer="{Binding MediaPlayer}">
    <MyOverlay/>
  </VideoView>
</Window>
```

without making the video pop out.

That got me looking into ways of being able to set the overlay inside the Window's axml as I didn't want to set it under VideoPlayer. Other thing that was frustrating was trying to inspect the tree using DevTools to debug my overlay. No content was showing up under VideoView (also the Window that VideoView creates was missing, but I didn't even know at the time that it even existed).

Thus I arrived at the Control that I submitted with this PR. It solved three problems. 

- I could set the overlay inside Window and bind the overlay's properties there
- I could replace the overlay without having to replicate VideoPlayer.axml but with different content, thus making VideoPlayer reusable
- I could use DevTools to inspect the overlay.

After I submitted this PR, I found out the reason behind those behaviors:

- Despite having a Content property, VideoView never sets it's LogicalChildren like ContentControl does ([https://github.com/AvaloniaUI/Avalonia/blob/master/src/Avalonia.Controls/ContentControl.cs]), so it never shows up in the inspector.

- VideoView runs under the assumption that either the MediaPlayer's setter or Initialized will always be called after CreateNativeControlCore(). However there is a case where CreateNativeControlCore() is called last. It means that under those conditions, the handle will never be passed down, causing the video to pop out on Windows and the app to crash silentely on Linux.

mfkl:
This is going in various directions but ok.. @winkmichael I don't have a proper bug report for this issue which I have not experience myself. Until then, it's impossible to get it fixed properly.

@nunofaria007 thanks for these investigations. Being able to capture events is a must, for things like DVD menus for example. I'm fine with providing a better integration than the current VideoView, but without breaking user compatibility ideally. We could maybe provide an additional control though.

nunofaria007:
Yeah, I'm sorry for mixing up a bunch of stuff. But after discovering the bug I don't think my PR matters anymore and I just wanted to alert for that issue.

Fixing the bug that makes the video pop out if VideoView is not part of a UserControl means that there's no need for a mandatory VideoPlayer Control anymore.

Best regards.

PR #423, opened 2026-04-10
Avalonia 12 support

mfkl:


src/LibVLCSharp.Avalonia/VideoView.cs
```diff
@@ -25,6 +26,65 @@ public class VideoView : NativeControlHost
...
+
+        // Avalonia 11: Visual.VisualRoot property (non-public)
+        private static readonly PropertyInfo? VisualRootProperty =
+            !IsAvalonia12OrAbove ? typeof(Visual).GetProperty("VisualRoot", BindingFlags.Public | BindingFlags.NonPublic | BindingFlags.Instance) : null;
+
+        // Avalonia 12: VisualExtensions.GetPresentationSource(visual).RootVisual
+        private static readonly MethodInfo? GetPresentationSourceMethod =
+            IsAvalonia12OrAbove ? typeof(global::Avalonia.VisualTree.VisualExtensions).GetMethod("GetPresentationSource", BindingFlags.Public | BindingFlags.Static, null, new[] { typeof(Visual) }, null) : null;
+        private static readonly PropertyInfo? RootVisualProperty =
+            GetPresentationSourceMethod?.ReturnType.GetProperty("RootVisual");
```
maxkatz6: `TopLevel.GetTopLevel(this) as Window` will work in both Avalonia 11 and 12

maxkatz6: `WindowDecorations`/`SystemDecorations` reflection is still likely needed, if you want to support both. But that's pretty simple compared to VisualRoot one.

Just make sure to apply trimmer attributes and test with AOT - otherwise compiled/trimmer might scream with warnings.

mfkl: I do want to support both, thanks for your review, will try it!

RobG66:
Tested this on windows ok.

PR #428, opened 2026-06-02
split uwp and winui from main package

mfkl:
fix https://code.videolan.org/videolan/LibVLCSharp/-/work_items/652

/cc @huynhsontung


huynhsontung:
Confirmed my project was built successfully with the new UWP package. I was trying to add just one extra package, but I couldn't find a way to split cleanly. This is way better. Thanks 👍

mfkl:
Thank you for your feedback @huynhsontung I will do a bit more testing/cleanup and merge/release this soon.

mfkl:
It does build, but did you try running your app?

huynhsontung:
I can't test that yet because there is outstanding work need to be done for Screenbox to run in UWP Native AOT. Right now, the app just crashes when I try to run it because of other issues unrelated to LibVLC. I can create a test application to validate this if you'd like.

mfkl:
No its all good, I did. But the UWP libvlc nuget needs updating I think, and that will likely break consumers. But you're using your own so you're good I think

mfkl:
Is your app still relying on 
```
<ItemGroup>
    <SDKReference Include="Microsoft.VCLibs, Version=14.0">
      <Name>Visual C++ 2015-2019 Runtime for Universal Windows Platform Apps</Name>
    </SDKReference>
  </ItemGroup>
```
?

PR #371, opened 2025-01-06
Avalonia: fix exception thrown when UserControl is unloaded

mfkl:
early return when visualroot is null.

bug report on SO https://stackoverflow.com/questions/79321503/how-to-navigate-between-usercontrols-while-using-vlc-videoview-with-content

PR #429, opened 2026-06-03
Avalonia: fix exception thrown when UserControl is unloaded

ibram-reda:
### Description of Change ###
i have create a test project `LibVLCSharp.Avalonia.Tests` which reference and test `LibVLCSharp.Avalonia` project
i have added only two tests to assure of `VideoView` removal correctly when it has a content or without a content

```c#
[TestCase(true)]
[TestCase(false)]
[AvaloniaTest]
public void VideoView_Should_Be_Removed_From_VisualTree_When_Parent_Is_Removed(bool withContent)
{
    // Setup controls:
    var videoView = new VideoView
    {
        Name = "VideoView",
        Content = withContent ? new TextBox() { Name = "VideoViewTextBox", Text = "" } : null
    };
    var secondView = new TextBox() { Name = "Second", Text = "" };
    var window = new Window { Content = videoView };
    // Open window:
    window.Show();
    try
    {                
        // change Content of window:
        window.Content = secondView;
        // wait for the UI to process the change:
        window.Dispatcher.RunJobs();
    }
    catch
    {
        // fail the test if any exception occurs during the process
        Assert.Fail("An exception occurred while changing the window content.");
    }
    var RemovedVideoView = window.FindDescendantOfType<VideoView>(false, t => t.Name == "VideoView");
    var secondTextBox = window.FindDescendantOfType<TextBox>(false, t => t.Name == "Second");
    Assert.That(RemovedVideoView, Is.Null);
    Assert.That(secondTextBox, Is.Not.Null);
}
```
this test was pass in contentless VedioView and Fail in the other one .... in order to make it pass in both cases i have added a guard before the problem line `PointToScreen`:
```c#
 if (!this.IsAttachedToVisualTree())
 {
       return;
 }

var newPosition = this.PointToScreen(topLeft);
```

this test should be the same test for the problem in https://github.com/videolan/libvlcsharp/pull/371

the other files is just add to make a test environment  Taken from [Avalonia docs : Headless Testing with NUnit](https://docs.avaloniaui.net/docs/testing/headless-nunit)
- app.axaml
- app.axaml.cs
- TestAppBuilder.cs

### DIFF
--- a/buildsystem/build.cake
+++ b/buildsystem/build.cake
@@ -26,7 +26,11 @@ var libraryProjects = new string[]
     "../src/LibVLCSharp.WinForms/LibVLCSharp.WinForms.csproj",
     "../src/LibVLCSharp.WPF/LibVLCSharp.WPF.csproj",
 };
-var testCsproj = "../src/LibVLCSharp.Tests/LibVLCSharp.Tests.csproj";
+var testCsprojs = new string[]
+{
+    "../src/LibVLCSharp.Tests/LibVLCSharp.Tests.csproj",
+    "../src/LibVLCSharp.Avalonia.Tests/LibVLCSharp.Avalonia.Tests.csproj",
+};
 
 var packagesDir = "../packages";
 var isCiBuild = BuildSystem.AzurePipelines.IsRunningOnAzurePipelines;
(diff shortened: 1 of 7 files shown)


mfkl:
Thanks! Will review and test in coming days.

mfkl:
Thanks!

PR #430, opened 2026-06-11
Fix inverted return value in Media.Statistics property

Krish2882005:
Swapped `default `and `mediaStats `to correctly return the populated struct.


mfkl:
Good catch, thanks!

PR #437, opened 2026-07-19
fix: TypeLoadException on ARM64 by replacing vsnprintf with _vscprintf

huynhsontung:
### Description of Change ###

Replaced the P/Invoke targeting `vsnprintf` in `msvcrt.dll` with `_vscprintf` for calculating required buffer lengths during logging on Windows platforms. 

### Rationale for Change
When running LibVLCSharp on ARM64 Windows devices, developers experience a constant stream of `System.TypeLoadException` warnings in their diagnostic logs (`Unresolved P/Invoke method 'msvcrt!vsnprintf_windows'`). While this doesn't crash the application, it creates significant noise and prevents LibVLC logs from being properly formatted and dispatched on ARM64.

Historically, `msvcrt.dll` did not export the standard C99 `vsnprintf` function. While modern x86 and x64 builds of Windows silently patched `msvcrt.dll` to include a `vsnprintf` forwarder (resolving to the modern UCRT), Microsoft did not include this forwarder in the ARM64 version of `msvcrt.dll`. 

Because `vsnprintf` is physically missing on ARM64 `msvcrt.dll`, the runtime fails to resolve the entry point, causing the noisy first-chance exceptions. However, `msvcrt.dll` *does* universally export `_vscprintf` (a Microsoft-specific function explicitly designed to count characters without writing them) across all architectures. By switching to `_vscprintf` to calculate the required buffer size, we avoid the missing entry point on ARM64 while retaining 100% compatibility across all legacy and modern Windows platforms.

### Issues Resolved ### 

- fixes #

### API Changes ###

 
 None. No public-facing API signatures or contracts were altered. The change is entirely internal to `LibVLCSharp.Shared.Helpers.MarshalUtils`.

### Platforms Affected ### 

- UWP

* **ARM64 Windows:** `GetLogMessage` now successfully formats and dispatches log messages. The application diagnostic logs are no longer flooded with `TypeLoadException` warnings caused by unresolved P/Invokes.
* **x86 / x64 Windows:** No functional change. `_vscprintf` calculates the buffer size identically to how `vsnprintf(IntPtr.Zero, UIntPtr.Zero, ...)` behaved previously.

Not applicable

### PR Checklist ###

### DIFF
--- a/src/LibVLCSharp/Shared/Helpers/MarshalUtils.cs
+++ b/src/LibVLCSharp/Shared/Helpers/MarshalUtils.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Collections.Generic;
 #if NET6_0_OR_GREATER
 using System.Diagnostics.CodeAnalysis;
@@ -61,8 +61,8 @@ namespace LibVLCSharp.Shared.Helpers
             [DllImport(Constants.Libc, EntryPoint = "vsnprintf", CallingConvention = CallingConvention.Cdecl)]
             public static extern int vsnprintf_linux(IntPtr buffer, UIntPtr size, IntPtr format, IntPtr args);
 
-            [DllImport(Constants.Msvcrt, EntryPoint = "vsnprintf", CallingConvention = CallingConvention.Cdecl)]
-            public static extern int vsnprintf_windows(IntPtr buffer, UIntPtr size, IntPtr format, IntPtr args);
+            [DllImport(Constants.Msvcrt, EntryPoint = "_vscprintf", CallingConvention = CallingConvention.Cdecl)]
+            public static extern int vscprintf_windows(IntPtr format, IntPtr args);
 #pragma warning restore IDE1006 // Naming Styles
         }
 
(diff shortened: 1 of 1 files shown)


huynhsontung:
This did not fix the issue. The NET Native compiler still complains that the method `_vscprintf`/`vscprintf` is unresolved and probably not allowed in UWP. Perhaps this issue has always been there; I just didn't notice it in Debug mode. The NET Native compiler doesn't like `msvcrt`.

mfkl:
Do you have an ARM64 device to test?

mfkl:
https://github.com/videolan/libvlcsharp/commit/0dfa493a9e234cda3ae194c6cb4de6816e6b394d can you test this on arm64?

huynhsontung:
> Do you have an ARM64 device to test?

Yes I do

huynhsontung:
> [0dfa493](https://github.com/videolan/libvlcsharp/commit/0dfa493a9e234cda3ae194c6cb4de6816e6b394d) can you test this on arm64?

That fixed it!

mfkl:
Ok will merge the fix, closing this then. Thanks for reporting and testing!

PR #441, opened 2026-07-21
Make LibVLC disposal idempotent

mfkl:


fallenmi (approved):
Reviewed with OpenAI Codex assistance; I independently reproduced the behavior against exact head `becf81e3fb4cf7ee8679541f1d69f6508347a97e`.

Looks good. In an isolated child process against real LibVLC, exact base exits 139 on the second `LibVLC.Dispose()`, while exact head exits cleanly and leaves `NativeReference == IntPtr.Zero`. The submitted regression and a separate equivalent oracle both pass on the head, including five additional isolated runs of each. The existing native-release and dialog-handler cleanup tests also pass, and the exact base/head test projects both build for `net10.0`.

The three upstream platform build jobs are green. The Azure test job was canceled at its 60-minute limit with only generic cancellation annotations, so it should still be rerun before merge; my approval is based on the exact local reproduction above rather than a green upstream test run.

mfkl:
@fallenmi this kind of AI comments which add no value are not welcome.

PR #442, opened 2026-07-21
Fixed trace.assert not accounting for hwndSource

NoahSimcox:
### Description of Change ###

LibVLCSharp.WPF's Foreground window class tries to access the parent window in order to apply foreground elements. A Trace.Assert() fails if GetWindow() is null but a wpf is not necessarily in a Window, it could be hosted in a HwndSource. The changes account for this change and only fail the assertion if the parent is neither a Window or a HwndSource.

### Issues Resolved ### 

No existing issue. Fixes a crash when LibVLCSharp.WPF is hosted in a HwndSource instead of a Window.

### API Changes ###

 
 None

### Platforms Affected ### 

- WPF

None

Not applicable

### Testing Procedure ###

1. Host a LibVLCSharp.WPF VideoView inside a WPF application where the control is hosted by an HwndSource instead of a Window.
2. Initialize the VideoView and verify that ForegroundWindow no longer triggers an assertion failure.
3. Verify that normal WPF Window hosting behavior remains unchanged.
4. Verify that video playback and foreground element handling continue to work as expected.

### PR Checklist ###

### DIFF
--- a/src/LibVLCSharp.WPF/ForegroundWindow.cs
+++ b/src/LibVLCSharp.WPF/ForegroundWindow.cs
@@ -76,9 +76,17 @@ namespace LibVLCSharp.WPF
             }
 
             _wndhost = GetWindow(_bckgnd);
-            Trace.Assert(_wndhost != null);
+
             if (_wndhost == null)
             {
+                if (PresentationSource.FromVisual(_bckgnd) is HwndSource)
+                {
+                    // Possibly Hosted WPF in a non-window source.
+                    // Therefore overlay would not be supported.
+                    return;
+                }
+
+                Trace.Assert(false, "VideoView is not hosted in a Window or HwndSource.");
                 return;
             }
 


mfkl:
hi, thanks for the PR. can you provide a small git repository with a sample that I can reproduce the issue with?

NoahSimcox:
yeah ill do that

NoahSimcox:
https://github.com/NoahSimcox/LibVLCSharp.WPF.ForegroundWindow_Test

NoahSimcox:
you'll notice that LibVLCSharp and LibVLCSharp.WPF are on 3.9.2 this is just because when i originally encountered this issue those were the versions I was using.

mfkl:
this does not build `  D:\a\1\s\src\LibVLCSharp.WPF\ForegroundWindow.cs(82,63): error CS0246: The type or namespace name 'HwndSource' could not be found (are you missing a using directive or an assembly reference?) [D:\a\1\s\src\LibVLCSharp.WPF\LibVLCSharp.WPF_npie32pg_wpftmp.csproj::TargetFramework=net6.0-windows]`

But whats your actual use case for this? I fail to see the point in your sample.

Also, silently disabling the foreground window will be very confusing for users and there is no documentation to explain why that is the case here.

I'm pretty much against merging this, unless you have some actual plausible use case for it

NoahSimcox:
The real problem here is that if a parent window is a HwndSource then ForegroundWindow essentailly can't handle it at all. The video still renders but becomes unwieldy, as important Foreground code can't run. The case in which the parent window would be a HwndSource would be if you had WPF code embedded in an older unmanaged C++ windows app. This is initially where I encountered this problem. 

To be honest it is probably best to not merge the current change as I now realize that for proper functionality ForegroundWindow would need to actaully handle a HwndSource as the parent window instead of what is proposed in the merge which is simply supressing the assertion failure. I might make this an issue but it could be more work than its worth.

For my own use case I can try to work around this problem, but I need the rendered video to have WPF elements overlayed on it which was the job of ForegroundWindow if im not mistaken. Either way don't bother with this current merge anymore.

mfkl:
Ok got you. Yeah, the foreground thing is kind of the best hack we can get for WPF at this time.

However, if you're willing to try libvlc/libvlcsharp 4 (see readme for preview builds), https://github.com/videolan/libvlcsharp/tree/master/samples/LibVLCSharp.CustomRendering.Direct3D11 this could work well on WPF for you.

Good luck!

PR #365, opened 2024-11-05
Fix Avalonia VideoView failing to update Content position when shifted.

Odalith:
### Description of Change ###

Adds an event handler for `VideoView.VisualRoot.LayoutUpdated `in `InitializeNativeOverlay()`.
Remove event handler for `BoundsProperty.Changed` as it's no longer needed.
Adds to the guard clause in `UpdateOverlayPosition()` so LayoutUpdated doesn't pelt a IsVisible = false VideoView with updates; Layout is updated on being made visible.

### Issues Resolved ### 
[Fixes 645](https://code.videolan.org/videolan/LibVLCSharp/-/issues/645)

### API Changes ###
 None

### Platforms Affected ### 
- Avalonia Desktop

### Behavioral/Visual Changes ###
VideoView Content now moves correctly when being shifted by grid splitters.

### Before/After Screenshots ### 

Before = bottom, After = top

### Testing Procedure ###
All testing was performed on Manjaro Linux.
- Tested with and without video playback.
- Tested behavior while VideoView.IsVisible = false.
- Tested changing VideoView.Content.
The tests where performed using [this](https://code.videolan.org/Odalith/vlctest/-/tree/master?ref_type=heads) application;

### PR Checklist ###

- [-] Rebased on top of the target branch at time of PR
- [-] Changes adhere to coding standard

PR #445, opened 2026-08-04
Add macOS ARM64 support

AnthonyKwon:
### Description of Change ###

Added Apple Silicon Mac support as Windows-style path structure.  
It still works with current `VideoLAN.LibVLC.Mac` package on x86, but it needs patched package with path structure changed and ARM64 files added to work on Apple Silicon.
PoC implementation of libvlc-nuget is available on [AnthonyKwon/libvlc-nuget](https://github.com/AnthonyKwon/libvlc-nuget). (It's not ready to be upstreamed yet as it uses hacky method to package)

### Issues Resolved ### 

- fixes [#363](https://code.videolan.org/videolan/LibVLCSharp/-/work_items/363)
- fixes [#365](https://code.videolan.org/videolan/LibVLCSharp/-/work_items/365)

### API Changes ###

 
 - `internal const string ArchitectureNames.MacOSArm64` in `Constants.cs`

### Platforms Affected ### 

- macOS

None

Not applicable

### Testing Procedure ###

- build this fork as standard LibVLCSharp macOS build procedure with all dependencies
- or, trim every other mobile and tvOS targets and its dependencies except `net9.0-macos` and `net10.0-macos` from `TargetFramework` and build only with Xcode needed.
  Since you're going to test only with macOS target, other targets are not needed to be tested. 
  I used this method for testing since I was unable to successfully build on some mobile platform. (seems not relevant with these changes since I was not even able to build pure LibVLCSharp)

[build.webm](https://github.com/user-attachments/assets/8446eb4d-f68c-4ee2-bb1e-3fa54baed64d)

### PR Checklist ###

### DIFF
--- a/src/LibVLCSharp/Shared/Core/Core.cs
+++ b/src/LibVLCSharp/Shared/Core/Core.cs
@@ -78,9 +78,8 @@ namespace LibVLCSharp.Shared
             {
                 arch = RuntimeInformation.ProcessArchitecture switch
                 {
-                    Architecture.X64 => Path.Combine(ArchitectureNames.MacOS64, Constants.Lib),
-                    Architecture.Arm64 => Path.Combine(ArchitectureNames.MacOSArm64, Constants.Lib),
-                    _ => Path.Combine(ArchitectureNames.MacOS64, Constants.Lib),
+                    Architecture.Arm64 => ArchitectureNames.MacOSArm64,
+                    _ => ArchitectureNames.MacOS64,
                 };
             }
             else if (PlatformHelper.IsWindows)
@@ -142,7 +141,7 @@ namespace LibVLCSharp.Shared
             paths.Add((string.Empty, libvlcPath3));
 
             // Add x64 folders as fallback for ARM64 to keep compatibility
-            if (arch == ArchitectureNames.WinArm64 || arch == Path.Combine(ArchitectureNames.MacOSArm64, Constants.Lib))
+            if (arch == ArchitectureNames.WinArm64 || arch == ArchitectureNames.MacOSArm64)
             {
                 var fallbackArchitecture = arch == ArchitectureNames.WinArm64
                     ? ArchitectureNames.Win64 : ArchitectureNames.MacOS64;
@@ -167,9 +166,8 @@ namespace LibVLCSharp.Shared
 
             if (PlatformHelper.IsMac)
             {
-                var libvlcPath4 = Path.Combine(Path.Combine(Path.GetDirectoryName(libvlcAssemblyLocation)!,
-                    Constants.Lib), $"{Constants.LibVLC}{LibraryExtension}");
-                var libvlccorePath4 = LibVLCCorePath(Path.Combine(Path.GetDirectoryName(libvlcAssemblyLocation)!, Constants.Lib));
+                var libvlcPath4 = Path.Combine(Path.GetDirectoryName(libvlcAssemblyLocation)!, $"{Constants.LibVLC}{LibraryExtension}");
+                var libvlccorePath4 = LibVLCCorePath(Path.GetDirectoryName(libvlcAssemblyLocation)!);
                 paths.Add((libvlccorePath4, libvlcPath4));
             }
 
--- a/src/LibVLCSharp/Shared/Core/Core.cs
+++ b/src/LibVLCSharp/Shared/Core/Core.cs
@@ -194,6 +194,19 @@ namespace LibVLCSharp.Shared
                 loadResult = LoadNativeLibrary(libvlcPath, out LibvlcHandle);
                 if (!loadResult)
                     Log($"Failed to load required native libraries at {libvlcPath}");
+                
+#if NET5_0_OR_GREATER
+                // register custom resolver to load library from provided path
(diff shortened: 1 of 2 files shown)


mfkl:
Thanks, code looks clean. This could be a single commit, you can fold them.

I'd like to see the build procedure for libvlc arm64 mac though.

And it'd be nice to make sure we still support mono and netcore targets, this was an issue when testing https://github.com/videolan/libvlcsharp/pull/163

AnthonyKwon:
I tried mono and netcore sample and it was failed for another reason. mono just simply throw `no such file or directory` so I need to see what's going on more, but I'm sure netcore doesn't support Apple Silicon since it throw `bad CPU type`.
I think it's because they're using old .NET version, so they lacks ARM64 support.
I'll upload video of building my fork of `libvlc-nuget` soon.

AnthonyKwon:
It downloads Intel and Apple Silicon variant of VLC, extracts it with 7z, and repackages it as correct structure.
yeah. really hacky method since I only considered personal usage when I first forked it. it needs much more care 😅

[nuget-build.webm](https://github.com/user-attachments/assets/882f4888-926b-48fd-871a-4c60d3603520)

mfkl:
that build is GPL also, so you're breaking the license unless your app is opensource. Hence why a proper libvlckit build is needed :)

AnthonyKwon:
yeah, my personal project is licensed with GPL so it's fine with it, but it needs to be split as LGPL/GPL variant like Windows if it's gonna actually used. but I think there might be alternative method.

mfkl:
the current vlckit nuget build is static (1 binary), the one from VLC is not (hundreds of plugins). This will break existing users. So lets make a static build of latest vlckit, for both x64 and arm, for 3.x branch, here on github.

mfkl:
or accept the breakage and publish both builds (x64 and arm) as dynamic

AnthonyKwon:
I made it work with statically built libvlc with new path scheme. but since this change might introduce real breakage which has possibility to break other OS, it needs more discussion before gets pushed here. (touches library loading)
I'll commit it later since current code is just hard-coded mess to just run demo for verifying my theory.

image

AnthonyKwon:
I have no idea about vlckit since it's new for me. I even never tried to build until you said it :(
at least (improved version of) workaround above seems make it load. I'll see what can I do for it.

AnthonyKwon:
seems like vlckit already can produce universal binary. it only needs sdk version set to higher one.
I'll leave detail separately on vlckit side. if universal binary is gonna be used, this PR no longer needed since that binary already runs with current code :)

src/LibVLCSharp/Shared/Core/Core.cs
```diff
@@ -197,8 +216,21 @@ static void LoadLibVLC(string? libvlcDirectoryPath = null)
             {
                 LoadNativeLibrary(libvlccore, out LibvlccoreHandle);
                 var loadResult = LoadNativeLibrary(libvlc, out LibvlcHandle);
-                if (loadResult)
-                    break;
+                if (!loadResult) continue;
+                
+#if NET5_0_OR_GREATER
+                // register custom resolver to load library from discovered path
+                NativeLibrary.SetDllImportResolver(Assembly.GetExecutingAssembly(), (dll, _, _) =>
```
mfkl, line 221: why make this call right after LoadNativeLibrary succeeded? This looks off.

AnthonyKwon: to filter out working path. I know it feels off, but when I looked for the P/Invoke exception causes (with inserting debug lines), it was coming from `DllImport` of `LibVLCVersion()`, and macOS seems to be have no problem at `DlOpen()` the LibVLC without path change, but calling any method throws error.
maybe I can add `&& MAC` to condition if you're unsure about it :)

AnthonyKwon: and this is why I'm calling it "workaround". maybe someone else who has better knowledge at low-level can bring better way to do it

mfkl: ok. but first as I said. libvlc build is the initial step.

AnthonyKwon:
I opened libvlc-nuget PR with static build on mfkl/libvlc-nuget#59.

src/LibVLCSharp/Shared/Core/Core.cs
```diff
@@ -188,6 +192,19 @@ static void LoadLibVLC(string? libvlcDirectoryPath = null)
                 loadResult = LoadNativeLibrary(libvlcPath, out LibvlcHandle);
                 if (!loadResult)
                     Log($"Failed to load required native libraries at {libvlcPath}");
+                
+#if NET5_0_OR_GREATER
+                // register custom resolver to load library from provided path
+                NativeLibrary.SetDllImportResolver(Assembly.GetExecutingAssembly(), (dll, _, _) =>
```
mfkl, line 198: SetDllImportResolver this changes the logic for all platforms target net5.0 and above. Remove it.

src/LibVLCSharp/Shared/Core/Core.cs
mfkl: SetDllImportResolver this changes the logic for all platforms target net5.0 and above. Remove it.

